PR #5, opened 2015-07-17
[LAB] ES6-ify

danielbreves:
:koala:

/cc @zendesk/quokka
### Risks
- TBD


lib/client.js
```diff
@@ -1,22 +1,22 @@
-var version      = require('version'),
-    Promise      = require('../vendor/ayepromise'),
+var _Promise     = require('../vendor/ayepromise'),
```
svizzari: What was the problem with es6/Babel `Promise`?

lib/client.js
```diff
@@ -36,14 +36,14 @@ Client.prototype = {
   /// var client = ZAFClient.init();
   /// client.postMessage('hello', { awesome: true });
   /// ```
-  postMessage: function(name, data) {
+  postMessage(name, data) {
     if (this.ready || name === 'iframe.handshake') {
       var msg = JSON.stringify({ key: name, message: data, appGuid: this._appGuid });
```
svizzari: Make this a `let`

lib/client.js
```diff
@@ -94,11 +94,11 @@ Client.prototype = {
   ///   client.off('app.activated', appActivated);
   /// });
   /// ```
-  off: function(name, handler) {
+  off(name, handler) {
     if (!this._messageHandlers[name]) { return false; }
     var index = this._messageHandlers[name].indexOf(handler);
```
svizzari: `let`?

lib/client.js
```diff
@@ -173,9 +173,9 @@ Client.prototype = {
   ///   }
   /// );
   /// ```
-  request: function(options) {
+  request(options) {
     var requestKey = 'request:' + requestCount++,
```
svizzari: `let`? :smile_cat:

svizzari:
@danielbreves This looks cool!  I stopped commenting on a few instances of `var` that could be `let`, but otherwise, :+1:

lib/client.js
danielbreves: It was 10 times the size and bigger than the SDK itself :/

svizzari: But it worked? Meh, `Promise` has [pretty wide support, except for IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Browser_compatibility), it's a shame we can't use it and rely on browser availability :crying_cat_face:  Babel can transpile and optionally exclude certain features (if they're already available in target browsers), right?

svizzari:
@sandlerr @ocke Give this some love if you get a spare minute, es6-ify everything!

lib/client.js
danielbreves: Yea, it worked :)

sandlerr:
is it possible to add a test case that the compiled artifact doesn't contain stuff that browsers don't support?

svizzari:
> is it possible to add a test case that the compiled artifact doesn't contain stuff that browsers don't support?

Testing whether babel works?  All of our target browser versions have ES5 support, right?

ocke:
ie9+ is es5, so yes.

sandlerr:
Testing that we've configured babel correctly. I know people who have accidentally deployed ES6 to the web because the pipeline wasn't babelify a particular file, with bad results for IE9 users.

danielbreves:
Closing in favour of https://github.com/zendesk/zendesk_app_framework_sdk/pull/75

PR #6, opened 2015-11-05
Use app.registered instead of app.activated

danielbreves:
:koala:

The docs included examples of listening to the `app.activated` event. This was confusing for app developers, because by the time the iframe is loaded the `app.activated` event has already fired. This PR changes examples to use `app.registered` instead, which is fired by the framework once the iframe is registered with the app.

/cc @zendesk/quokka

### Risks
 - None


sandlerr:
:+1:

svizzari:
:+1:

danielbreves:
cc @chucknado

adammw:
:+1:

PR #8, opened 2016-02-05
[AF-216] [PoC] Initial exploration of new APIs for the App Framework SDK

adammw:
:v: :construction: 

:no_entry: **See TODO section** :no_entry: 

/cc @zendesk/vegemite

### Description
See https://github.com/zendesk/zendesk_app_framework/pull/537.

### To do
- [ ] Ensure this is backwards compatible to avoid breaking apps using /latest/zaf_sdk.js

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-216

### Risks
**TODO**


lib/client.js
```diff
@@ -192,6 +246,97 @@ Client.prototype = {
...
+  ///
+  /// ```javascript
+  /// var client = ZAFClient.init();
+  ///
+  /// client.metadata().then(function(metadata) {
+  ///   console.log(metadata); // { appId: 1234, installationId: 1234 }
+  /// });
+  /// ```
+  metadata: function() {
+    var deferred = Promise.defer();
```
adammw: cache the metadata promise

lib/client.js
```diff
@@ -1,19 +1,85 @@
...
+function isValidEvent(client, event) {
+  return client._origin === event.origin && client._source === event.source;
+}
+
+function messageHandler(client, event) {
+  if (!isValidEvent(client, event)) { return; }
+
+  var data = event.data;
+
+  if (!data) { return; }
```
sandlerr: do we always want that?

lib/client.js
```diff
@@ -1,19 +1,85 @@
...
+        err.name = data.error.code;
+        err.stack = data.error.stack;
+      }
+      pendingPromise.reject(err);
+    } else {
+      pendingPromise.resolve(data.result);
+    }
+  } else if (ZAF_EVENT.test(data.key)) {
+    var key = data.key.replace(ZAF_EVENT, '');
+    client && client.trigger(key, data.message);
```
sandlerr: this is backward compatibility, right?

lib/client.js
```diff
@@ -192,6 +254,172 @@ Client.prototype = {
...
+  /// A [Promises/A+](https://promisesaplus.com) conformant `promise` object.
+  ///
+  /// ```javascript
+  /// var client = ZAFClient.init();
+  ///
+  /// client.call('modal.show', { url: 'https://www.example.com/modal.html' }).then(function(data) {
+  ///   console.log(data); // { width: 200, height: 200 }
+  /// });
+  /// ```
+  call: function(name) {
```
sandlerr: let's not call it `call` since we aren't using falcor

lib/client.js
```diff
@@ -192,6 +254,172 @@ Client.prototype = {
...
+    deferred.promise.then(removePromise.bind(null, id), removePromise.bind(null, id));
+
+    // timeout the promise to ensure it will be garbage collected if nobody responds
+    setTimeout(function() {
+      deferred.reject(new Error('UI Request Timeout'));
+    }, PROMISE_TIMEOUT);
+
+    pendingPromises[id] = deferred;
+
+    var msg = JSON.stringify({ id: id, request: name, arguments: args, appGuid: this._appGuid });
```
sandlerr: I'd prefer to avoid calling things `arugments` as well

lib/client.js
adammw: got some suggestions? call does fit the best in my mind at least, but I do see how it could be confused with Function.prototype.call

lib/client.js
adammw: I think so, it was in the previous message handler too. Basically we need to filter out any postMessages that aren't in the format we expect (i.e. from other apps/sdks/etc).

lib/client.js
adammw: @sandlerr you want to remove events? or just restructure the message format for them?

sandlerr: I'm not suggesting any changes, just trying to understand the code

lib/client.js
sandlerr: exec?

svizzari: What about `perform` or `execute`?

lib/client.js
svizzari: `params` could work?

lib/client.js
danielbreves: `send` or `action`?

lib/client.js
```diff
@@ -192,6 +254,172 @@ Client.prototype = {
...
+  ///   * `paths` an array of path strings or a single path string
+  ///
+  /// ##### Returns
+  ///
+  /// A [Promises/A+](https://promisesaplus.com) conformant `promise` object.
+  ///
+  /// ```javascript
+  /// var client = ZAFClient.init();
+  ///
+  /// client.ui(['ticket.subject', 'ticket.fields.custom_field_1234.value').then(function(data) {
```
svizzari: Doc should be `get`, not `ui`.  Also, missing a closing square bracket after the `'ticket.fields.custom_field_1234.value'`

lib/client.js
```diff
@@ -1,19 +1,108 @@
...
+
+  if (typeof data === 'string') {
+    try {
+      data = JSON.parse(event.data);
+    } catch (e) {
+      return;
+    }
+  }
+
+  console.debug('SDK <<', data);
```
ocke: Do we need this in production?

ocke:
I'm pretty happy with all of this (except the console statement?).

👍

lib/client.js
danielbreves: Probably not, we could use something like Lotus' minilog, where it only logs when enabled. I wouldn't oppose to removing it for now.

spec/client_spec.js
```diff
@@ -173,6 +231,33 @@ describe('Client', function() {
 
     });
 
+    describe('#get', function() {
+      beforeEach(function() {
+        sandbox.spy(window, 'setTimeout');
+      });
+
+      it('returns a promise', function() {
+        expect(subject.get('ticket.subject')).to.be.a.promise;
```
danielbreves: Can we also test that this promise gets resolved when the expected message is received by the handler?

ocke: how?

danielbreves: It's a little hacky, but the way we've been triggering the message event is by [stubbing `window.addEventListener`](https://github.com/zendesk/zendesk_app_framework_sdk/blob/adammw/sdk-v2/spec/client_spec.js#L12) and then [calling the argument](https://github.com/zendesk/zendesk_app_framework_sdk/blob/adammw/sdk-v2/spec/client_spec.js#L68) it receives, which is the [`messageHandler`](messageHandler). This way you can pass an event in the format the sdk expects and test that it resolves the promise.

ocke:
:+1:

lib/client.js
```diff
@@ -1,19 +1,121 @@
...
+function cachedResponseToMessage(message, cacheName) {
+  var deferred = Promise.defer();
+  function resolveMetadata() {
+    if (this[cacheName]) {
+      deferred.resolve(this[cacheName]);
+    } else {
+      this.on(message, function() {
+        deferred.resolve(this[cacheName]);
+      }.bind(this));
+    }
```
adammw: move this conditional up one level

adammw: potentially add a test?

lib/client.js
```diff
@@ -1,19 +1,121 @@
...
   this._appGuid = appGuid;
   this._messageHandlers = {};
+  this._metadata = null;
+  this._context = null;
   this.ready = false;
 
-  this.on('app.registered', function() {
+  this.on('app.registered', function(data) {
     this.ready = true;
+    this._metadata = data.metadata;
```
adammw: this can be removed if we have cachedResponseToMessage doing it.

adammw:
:+1: - although do we want to continue to have `.request()` use the same old code path? We can always update it later...

sandlerr:
:+1:

ocke:
👍

PR #12, opened 2016-04-27
[AF-311] Revert docs that are now live

ocke:
:v:

/cc @zendesk/vegemite

### Description
Revert docs being live

### References
* JIRA: 

### Risks

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -278,21 +278,6 @@ Client.prototype = {
     return deferred.promise;
   },
 
-  /// #### client.metadata()
-  ///
-  /// Request metadata for the app, such as the app ID and installation ID.
-  ///
-  /// ##### Returns
-  ///
-  /// A [Promises/A+](https://promisesaplus.com) conformant `promise` object.
-  ///
-  /// ```javascript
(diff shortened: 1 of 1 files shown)


sandlerr:
:+1:

svizzari:
Retrospective 👍

sandlerr:
@ocke this needs a followup to put them in a group

danielbreves:
Why not keep them with a double slash instead?

sandlerr:
@ocke is going to revert this revert to make them available to a beta group in hercules. We found this was the only way to make a doc only visible to a group.

ocke:
I've reverted this

danielbreves:
@sandlerr so we got rid of these docs to unpublish them, now we revert the revert to publish them to the beta group?

danielbreves:
For that to work the file needs to be hidden, not just the diff right? That's not clear from the UI.

sandlerr:
yes that's right. @ocke is going to create a new 'v2' file for these docs. Any day now.

sandlerr:
Sorry, just saw that you're working on it

PR #14, opened 2016-05-04
Ability to pass what window we want to link with!

ocke:
:v:

/cc @zendesk/vegemite
### Description

ability to pass the source of the window

So insteaf of doing this:
`zafClient.init();`

you could do this to bind to a certain level of the framework. Which means you could potentially create apps in zopim that talk to the zopim layer AND the zendesk layer!
`zafClient.init( window.parent.parent );`
### References
- JIRA:

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -80,9 +80,9 @@ function messageHandler(client, event) {
   }
 }
 
-var Client = function(origin, appGuid) {
+var Client = function(origin, appGuid, source) {
   this._origin = origin;
-  this._source = window.parent;
+  this._source = source || window.parent;
   this._appGuid = appGuid;
   this._messageHandlers = {};
   this._metadata = null;
--- a/lib/client.js
+++ b/lib/client.js
@@ -82,7 +82,7 @@ function messageHandler(client, event) {
 
 var Client = function(origin, appGuid, source) {
   this._origin = origin;
-  this._source = source || window.parent;
+  this._source = typeof source === 'object' ? source : window.parent;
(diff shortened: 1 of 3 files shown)


sandlerr:
should we check if it's a function? What if someone wants to pass a source but no callback?

danielbreves:
Why do we need this?

ocke:
So you can create 2 clients, one for zopim and one for zendesk! Cool huh!

danielbreves:
Maybe, but we should validate use cases before writing code, otherwise it becomes bloatware. There are many other cool things to be built as well as nasty bugs to be fixed :)

ocke:
It's 3 lines of code...
The tests needed to be written anyway.

sandlerr:
specific usecase: the top bar zopim app now loads an iframe with the KB search app inside. If we had an API where you an app could open a user profile in Lotus, the KB search app would need to postMessage the top window to make that happen (it displays links to user profiles but they open in new tabs).

danielbreves:
I understand the use case, but we have yet to have customers using apps in Zopim, let alone apps in the Zopim app. Also, we've been told customers actually prefer using Zopim standalone on a different tab, even when they use both Zopim and Zendesk. It may make sense in the future, but I think it's unlikely this would get enough usage in the short term to justify the added complexity.

Besides, ZAF doesn't support multiple clients right now. If another handshake is posted ZAF [deregisters](https://github.com/zendesk/zendesk_app_framework/blob/master/lib/zaf_sdk_client.js#L36) the first client.

PR #16, opened 2016-05-23
Olaf/no defer

ocke:
:v:

/cc @zendesk/vegemite

### Description
We have used the old style of Promises. I replaced it with the new style so it works in all browsers.

More info: http://stackoverflow.com/questions/27889687/promise-defer-browser-support
### References
* JIRA: 

### Risks
- SDK stops working


bower.json
```diff
@@ -22,6 +22,6 @@
     "tests"
   ],
   "dependencies": {
-    "ayepromise": "~1.1.2"
+    "es6-promise": "^3.2.1"
```
svizzari: @danielbreves concern when we added ayepromise, was that es6-promise was way bigger than ayepromise, and bigger even than the SDK IIRC.

lib/client.js
```diff
@@ -1,7 +1,6 @@
 var PROMISE_TIMEOUT = 5000, // 5 seconds
     ZAF_EVENT       = /^zaf\./,
     version         = require('version'),
-    Promise         = window.Promise || require('../vendor/ayepromise'),
```
svizzari: `window.Promise` doesn't exist for IE11, right?

ocke: yeah i reverted... Roman just told me. I spent 30 min trying to get rid of it, but to no avail. Thanks MS!

bower.json
ocke: Correct.
But ayepromise doesn't support the new way of doing promises...

ocke: I had a quick look around and found this library called 'native-promise-only' and it's smaller and seems to work fine 👍

lib/client.js
```diff
@@ -15,29 +15,32 @@ function rawPostMessage(client, msg, forceReady) {
...
+      timeoutId,
+      params = Array.prototype.slice.call(arguments, 1);
 
-  // timeout the promise to ensure it will be garbage collected if nobody responds
-  var timeoutId = setTimeout(function() {
-    deferred.reject(new Error('Invocation request timeout'));
-  }, PROMISE_TIMEOUT);
+  var promise = new Promise(function(resolve, reject) {
+    // timeout the promise to ensure it will be garbage collected if nobody responds
+    // we don't need to reject, because if an error is thrown in a promise it automatically rejects.
```
sandlerr: I dont understand this line

spec/client_spec.js
```diff
@@ -226,17 +226,35 @@ describe('Client', function() {
...
-        sandbox.spy(window, 'setTimeout');
+        requestsCount++;
       });
 
       it('returns a promise', function() {
-        expect(subject.get('ticket.subject')).to.be.a.promise;
+        var promise = subject.get('ticket.subject');
+
+        promise.catch(function() {});
+        expect(promise).to.be.a.promise;
```
sandlerr: let's put this above the `catch`, that way if it isn't a promise we get a nice assertion failure message rather than a `called catch on undefined` error

spec/client_spec.js
```diff
@@ -226,17 +226,35 @@ describe('Client', function() {
...
       });
 
-      it('rejects the promise after 5 seconds', function() {
-        subject.get('ticket.subject');
-        expect(setTimeout).to.have.been.calledWith(sinon.match.func, 5000);
+      it('accepts an array with mutliple paths', function() {
```
sandlerr: spelling of `multiple`

spec/client_spec.js
```diff
@@ -249,22 +267,25 @@ describe('Client', function() {
...
-        ++requestsCount;
       });
     });
 
     describe('#set', function() {
       it('returns a promise', function() {
-        expect(subject.set('ticket.subject', 'value')).to.be.a.promise;
+        var promise = subject.set('ticket.subject', 'value');
+
+        promise.catch(function() {});
```
sandlerr: actually, we should put all these `catch` statements in an `afterEach` since they aren't part of the test

sandlerr:
good fix!

spec/client_spec.js
ocke: yeah i tried that, but then promise refers to the wrong promise, and i needed to make an exception for the one where we do what the timeout to run. :/

ocke:
is that a 👍 ?

sandlerr:
:+1:

lib/client.js
```diff
@@ -15,29 +15,31 @@ function rawPostMessage(client, msg, forceReady) {
...
-  return deferred.promise;
+  // ensure promise is cleaned up when resolved
+  return promise.then(removePromise.bind(null, id, timeoutId), removePromise.bind(null, id, timeoutId));
 }
 
-function removePromise(id, timeoutId) {
+function removePromise(id, timeoutId, args) {
   clearTimeout(timeoutId);
   delete pendingPromises[id];
+  if (args instanceof Error) throw args;
```
svizzari: This is so the error bubbles to any chained promise handlers in the event that the promise being removed fails?

public/index.html
```diff
@@ -2,32 +2,11 @@
...
-      function(args) {
-        console.log('success:', args);
-      },
-      function(args) {
-        console.log('fail:', args);
-      }
-    );
-  </script>
+  <script src="https://cdn.jsdelivr.net/g/jquery@2.2.3,underscorejs@1.8.3"></script>
+  <script src="http://localhost:5678/framework.js""></script>
```
svizzari: One too many double quotes here after framework.js

spec/client_spec.js
```diff
@@ -226,45 +229,77 @@ describe('Client', function() {
...
+
+        expect(promise).to.be.a.promise;
+      });
+
+      it('rejects the promise after 5 seconds', function(done) {
+        this.timeout(6000);
+
+        promise = subject.get('ticket.subject');
+
+        promise.catch(function(err) {
```
svizzari: We have Chai as Promised, right? Could possibly do something like: `expect(promise).should.be.rejectedWith(Error)`?

lib/client.js
sandlerr: Exactly, Olaf discovered that we had unintended behaviour by doing:

`promise.catch(myFunc).then(yourFunc).catch(yourOtherFunc)`

if the promise rejects, `myFunc` runs, but if `myFunc` doesn't `throw` then `yourFunc` runs. We intended for users to be able to `catch` themselves.

svizzari:
Looks pretty good.  👍 when my comment in the index.html file is addressed

public/index.html
ocke: good find. You scrutinise my code with a microscope!

spec/client_spec.js
ocke: I tried this, but it says 'cannot find rejectedWith on undefined' ? I looked in the docs, and you're doing it right. Not sure why :|

lib/client.js
ocke: yeah. If somebody writes code that errors, or themself throw an error, our 'in between .then', where we remove the promises from the array, i.e. removePromise (line #35) eats up those errors. This line makes sure it does get thrown and the person will see a nice error and/or can handle it themself. It's their error after all!

public/index.html
```diff
@@ -2,32 +2,11 @@
...
-        console.log('success:', args);
-      },
-      function(args) {
-        console.log('fail:', args);
-      }
-    );
-  </script>
+  <script src="https://cdn.jsdelivr.net/g/jquery@2.2.3,underscorejs@1.8.3"></script>
+  <script src="http://localhost:5678/framework.js"></script>
+  <iframe src="iframe.html?origin=http://localhost:9001&amp;app_guid=1234"></iframe>
```
danielbreves: @ocke how will this work without installed.js creating the app instance?

ocke: @danielbreves yeah it doesn't. I was hoping in the future it would work.

danielbreves: @ocke can we revert it to the way it was before, for now?

PR #19, opened 2016-06-10
[AF-292] As an App Developer, I want to register a save hook from my IFO app

svizzari:
✌️

/cc @zendesk/vegemite

### Description

Implementation of v2 save hook in the SDK.

Relies on a PR in ZAF, zendesk/zendesk_app_framework#573

### TODO
- [x] ~~Add tests for `when` utility function~~ a36767f

### References
* ZAF PR:  zendesk/zendesk_app_framework#573
* JIRA: https://zendesk.atlassian.net/browse/AF-292

### Risks
* [low] The triggering of custom events within the iframe might be adversely affected by the special-casing of ZAF events that expect a reply.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -3,8 +3,14 @@ var PROMISE_TIMEOUT = 5000, // 5 seconds
     version         = require('version'),
     Promise         = window.Promise || require('../vendor/native-promise-only'),
     pendingPromises = {},
-    promiseCount    = 0,
-    requestCount    = 0;
+    ids             = {};
+
+function nextIdFor(name) {
+  if (isNaN(ids[name])) {
+    ids[name] = 0;
+  }
+  return ++ids[name];
+}
 
 function rawPostMessage(client, msg, forceReady) {
   if (client.ready || forceReady) {
@@ -15,7 +21,7 @@ function rawPostMessage(client, msg, forceReady) {
 }
 
 function wrappedPostMessage(name) {
-  var id = ++promiseCount,
+  var id = nextIdFor('promise'),
       timeoutId,
       params = Array.prototype.slice.call(arguments, 1);
 
@@ -158,6 +164,8 @@ Client.prototype = {
   /// called immediately after the `init` callback.
   on: function(name, handler, context) {
     if (typeof handler == 'function') {
+      var id = nextIdFor('event');
+      
       this._messageHandlers[name] = this._messageHandlers[name] || [];
       this._messageHandlers[name].push(context ? handler.bind(context) : handler);
     }
@@ -260,7 +268,7 @@ Client.prototype = {
   /// );
   /// ```
   request: function(options) {
-    var requestKey = 'request:' + requestCount++;
+    var requestKey = 'request:' + nextIdFor('request');
 
     return new Promise(function(resolve, reject) {
       if (typeof options === 'string') {
--- a/lib/client.js
+++ b/lib/client.js
@@ -188,7 +188,7 @@ Client.prototype = {
       this._messageHandlers[name].push(handler);
 
       if (name !== 'app.registered') {
-        rawPostMessage(this, { key: 'iframe.on:' + name, id: id, appGuid: this._appGuid });
+        this.postMessage('iframe.on:' + name);
       }
     }
   },
@@ -213,6 +213,7 @@ Client.prototype = {
   off: function(name, handler) {
     if (!this._messageHandlers[name]) { return false; }
     var index = this._messageHandlers[name].indexOf(handler);
+    this.postMessage('iframe.off:' + name);
     return this._messageHandlers[name].splice(index, 1)[0];
   },
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -3,7 +3,6 @@ var PROMISE_TIMEOUT = 5000, // 5 seconds
     version         = require('version'),
     Promise         = window.Promise || require('../vendor/native-promise-only'),
     when            = require('utils').when,
-    repliesPending  = {},
     pendingPromises = {},
     ids             = {};
 
@@ -95,6 +94,7 @@ var Client = function(origin, appGuid) {
   this._source = window.parent;
   this._appGuid = appGuid;
   this._messageHandlers = {};
+  this._repliesPending = {};
   this._metadata = null;
   this._context = null;
   this.ready = false;
@@ -243,22 +243,15 @@ Client.prototype = {
   /// client.trigger('activation') // activating!
   /// ```
   trigger: function(name, data) {
-    if (!this._messageHandlers[name] || repliesPending[name]) { return false; }
+    if (!this._messageHandlers[name] || this._repliesPending[name]) { return false; }
     if (data && data.reply) {
-      repliesPending[name] = true;
-      // When receives a list of things to wrap as Promises,
-      // it will then wait for:
-      //  a) all to settle
-      //  b) first to reject
-      //  c) first to settle with a falsy value (?)
-      // c) is sub-optimal, since it has a specific narrow use-case
-      // for save hooks :(
+      this._repliesPending[name] = true;
       when(this._messageHandlers[name]).then(
         this.postMessage.bind(this, 'iframe.reply:' + name)
       ).catch(function(reason) {
         this.postMessage('iframe.reply:' + name, { error: true, reason: reason }); 
       }.bind(this)).then(
-        function() { delete repliesPending[name]; }
+        function() { delete this._repliesPending[name]; }
       );
     } else {
       this._messageHandlers[name].forEach(function(handler) {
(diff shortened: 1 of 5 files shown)


lib/client.js
```diff
@@ -229,10 +243,28 @@ Client.prototype = {
   /// client.trigger('activation') // activating!
   /// ```
   trigger: function(name, data) {
-    if (!this._messageHandlers[name]) { return false; }
-    this._messageHandlers[name].forEach(function(handler) {
-      handler(data);
-    });
+    if (!this._messageHandlers[name] || repliesPending[name]) { return false; }
+    if (data && data.reply) {
```
adammw: could potentially conflict with any user-defined `reply` property, perhaps move it up one level and allow a third parameter [here](https://github.com/zendesk/zendesk_app_framework_sdk/blob/6ecb1a2f78b8ea84e3680f3409b8ec7f60ae21c5/lib/client.js#L88)? Or as trigger is only ever called by the messageHandler itself for hooks, (we don't really want to support user-defined hooks do we), could potentially split this into a private method and leave trigger for the simple 'event' use case?

lib/client.js
```diff
@@ -229,10 +243,28 @@ Client.prototype = {
   /// client.trigger('activation') // activating!
   /// ```
   trigger: function(name, data) {
-    if (!this._messageHandlers[name]) { return false; }
-    this._messageHandlers[name].forEach(function(handler) {
-      handler(data);
-    });
+    if (!this._messageHandlers[name] || repliesPending[name]) { return false; }
```
adammw: why is one variable local (to the client) and the other global? Could we put them both on the client object?

lib/client.js
```diff
@@ -229,10 +243,28 @@ Client.prototype = {
...
+      //  a) all to settle
+      //  b) first to reject
+      //  c) first to settle with a falsy value (?)
+      // c) is sub-optimal, since it has a specific narrow use-case
+      // for save hooks :(
+      when(this._messageHandlers[name]).then(
+        this.postMessage.bind(this, 'iframe.reply:' + name)
+      ).catch(function(reason) {
+        this.postMessage('iframe.reply:' + name, { error: true, reason: reason }); 
+      }.bind(this)).then(
```
adammw: will this then be called for errors as well?

lib/client.js
```diff
@@ -229,10 +243,28 @@ Client.prototype = {
...
   /// ```
   trigger: function(name, data) {
-    if (!this._messageHandlers[name]) { return false; }
-    this._messageHandlers[name].forEach(function(handler) {
-      handler(data);
-    });
+    if (!this._messageHandlers[name] || repliesPending[name]) { return false; }
+    if (data && data.reply) {
+      repliesPending[name] = true;
+      // When receives a list of things to wrap as Promises,
```
adammw: I wonder if this comment should go near the when definition rather than it's usage

lib/client.js
svizzari: > Could we put them both on the client object?

Yes, we can do that.  If we're doing so, we should probably put the `pendingPromises` module-local var into the client to.  This would have the advantage of making it easier to allow multiple clients in the one app, (when/if we want to do that).

lib/client.js
adammw: I'm convinced... 
screen shot 2016-06-15 at 9 50 05 am

lib/utils.js
```diff
@@ -26,7 +28,82 @@ function queryParameters(queryString) {
...
+
+function isFalsy(val) {
+  return !val ||
+    val === false ||
+    val instanceof Error ||
+    typeof val === 'string';
+}
+
+// items should be a list
+function when(items) {
```
adammw: can we unit-test this bad-boy?

svizzari: sure can, will do

lib/client.js
svizzari: Yeah, I will move this comment to the utils module and elaborate

lib/utils.js
```diff
@@ -26,7 +28,93 @@ function queryParameters(queryString) {
...
 }
 
+function isPromise(obj) {
+  return obj instanceof Promise ||
+    !!obj && obj.then && typeof obj.then === 'function';
+}
+
+function isFalsy(val) {
+  return !val ||
+    val === false ||
```
danielbreves: Do we need this check when we're already checking `!val`?

danielbreves: Do we need this check when we're already checking `!val`?

lib/utils.js
```diff
@@ -26,7 +28,93 @@ function queryParameters(queryString) {
...
+    val instanceof Error ||
+    typeof val === 'string';
```
danielbreves: Why do we consider strings falsy?

lib/utils.js
```diff
@@ -26,7 +28,93 @@ function queryParameters(queryString) {
...
+
+// When receives a list of things to wrap as Promises.
+// it will then wait for:
+//  a) all to settle
+//  b) first to reject
+//  c) first to settle with a falsy value
+// c) is sub-optimal, since it has a specific narrow use-case
+// for save hooks
+function when(items) {
+  items = !items ? [] : items;
```
danielbreves: `items = items || [];`?

lib/utils.js
svizzari: Yes, when we're emulating the save hook behaviour.  A string returned indicates a custom error message.

lib/utils.js
ocke: haha, I love `items = !items ? [] : items`

lib/utils.js
ocke: `!val` already includes `val === false` so that test is irrelevant.

svizzari:
@danielbreves 💇 Should be 💚 in Travis now

danielbreves:
👍

lib/client.js
```diff
@@ -75,8 +82,22 @@ function messageHandler(client, event) {
       pendingPromise.resolve(data.result);
     }
   } else if (ZAF_EVENT.test(data.key)) {
-    var key = data.key.replace(ZAF_EVENT, '');
-    if (client) {
+    var key = data.key.replace(ZAF_EVENT, ''),
+        msg = { appGuid: client._appGuid };
+    if (data.reply) {
```
adammw: nitpick: `needsReply` would make more sense to me when reading, (so as not to get confused between the request and the reply itself), but no need to change since that would require a ZAF change as well.

lib/client.js
```diff
@@ -75,8 +82,22 @@ function messageHandler(client, event) {
       pendingPromise.resolve(data.result);
     }
   } else if (ZAF_EVENT.test(data.key)) {
-    var key = data.key.replace(ZAF_EVENT, '');
-    if (client) {
+    var key = data.key.replace(ZAF_EVENT, ''),
+        msg = { appGuid: client._appGuid };
```
adammw: nitpick: could define msg inside the reply if statement, and define key inline (it's not used in the else block)

lib/client.js
```diff
@@ -75,8 +82,22 @@ function messageHandler(client, event) {
...
+    if (data.reply) {
+      if (client._repliesPending[key]) { return; }
+      msg.key = 'iframe.reply:' + key;
+      client._repliesPending[key] = true;
+      when(client._messageHandlers[key]).then(
+        rawPostMessage.bind(null, client, msg)
+      ).catch(function(reason) {
+        msg.error = true;
```
adammw: could make this the same format as the error responses for promises, i.e. { error: { msg: 'reason' } } - although again, that would require a ZAF change.

adammw:
Just some nitpicks / suggestions that would require a zaf change... Feel free to leave them as is. 👍

lib/client.js
danielbreves: I found that confusing too.

svizzari: Ha, I originally used `needsReply` then decided `reply` was cleaner because simpler.  Can change it back

lib/client.js
svizzari: If I change ZAF for `needsReply` I'll do this too

svizzari:
@adammw @danielbreves 💇 See zendesk/zendesk_app_framework#579

danielbreves:
still 👍

adammw:
Is there a PR to change the error consumption?

svizzari:
> Is there a PR to change the error consumption?

Added to the `needsReply` PR to save on overhead (of creating another)

https://github.com/zendesk/zendesk_app_framework/pull/579/commits/121d4ef5a515f4ff1e3aa7e0b548376c725e5ca5

adammw:
:+1: (again)

lib/utils.js
```diff
@@ -26,7 +28,91 @@ function queryParameters(queryString) {
...
+          });
+        }
+      } catch(err) {
+        promise = new Promise(function(resolve, reject) {
+          reject(err);
+        });
+      }
+    } else {
+      promise = new Promise(function(resolve, reject) {
+        isFalsy(item) ?
```
adammw: @svizzari - I'm confused by this - was the intention to check item? as in if someone does `client.on('ticket.save', false)` ?

PR #20, opened 2016-06-14
sdk fixes

ocke:
:v:

/cc @zendesk/vegemite

### Description
Fxies some array in array wrapping
https://github.com/zendesk/zendesk_app_framework/pull/574 needs to be deployed first.

### References
* JIRA: 

### Risks

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -307,7 +307,9 @@ Client.prototype = {
   get: function(path) {
     var paths = Array.isArray(path) ? path : Array.prototype.slice.call(arguments, 0);
 
-    if (paths.filter(function(s) { return typeof s !== 'string'; }).length) { return Promise.reject('The get method accepts a string or array of strings.'); }
+    if (paths.some(function(s) {return typeof s !== 'string'; })) {
+      return Promise.reject('The get method accepts a string or array of strings.');
+    }
 
     return wrappedPostMessage.call(this, 'get', paths);
   },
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -14,10 +14,10 @@ function rawPostMessage(client, msg, forceReady) {
   }
 }
 
-function wrappedPostMessage(name) {
+/* params is an array for set/invoke or an object for set */
```
svizzari: How come this one line is a block comment?

ocke: because it starts with `/*` and ends with `*/`

danielbreves: did you mean `get` instead of `set`?

lib/client.js
```diff
@@ -14,10 +14,10 @@ function rawPostMessage(client, msg, forceReady) {
   }
 }
 
-function wrappedPostMessage(name) {
+// params is an array for get/invoke or an object for set
+function wrappedPostMessage(name, params) {
```
danielbreves: I realise that this is an existing issue, but I don't like the fact the `params` array represents multiple calls for `get` but a single call for `invoke`.

lib/client.js
```diff
@@ -303,22 +303,46 @@ Client.prototype = {
     }.bind(this));
   },
 
+  // Accepts string or array of strings.
   get: function(path) {
-    var paths = Array.isArray(path) ? path : [path];
+    var paths = Array.isArray(path) ? path : Array.prototype.slice.call(arguments, 0);
```
adammw: i'm skeptical

lib/client.js
```diff
@@ -303,22 +303,46 @@ Client.prototype = {
...
-  invoke: function() {
-    return wrappedPostMessage.bind(this, 'invoke').apply(null, arguments);
+  invoke: function(key) {
+    var obj = key;
+
+    if (typeof key === 'string') {
+      obj = {};
+      obj[key] = Array.prototype.slice.call(arguments, 1);
+    } else {
+      throw new Error('Invoke with an object isn\'t supported (yet).');
```
adammw: let's not say yet, that implies that we will ever get around to fixing it

adammw:
I'd prefer to leave out the multiple argument conversion into an array for the get, seems a bit too magic for our public SDK and limits us if we want to add other arguments in the future that do different things. It's shouldn't be that hard for developers to just create and use an array. 
That being said, I can be swayed if the rest of the team wants this.

adammw:
:+1:

danielbreves:
👍  but let's not merge this until zendesk/zendesk_app_framework#574 is in production.

ocke:
but it will only create beta4?

danielbreves:
ah right, I guess it's ok then

ocke:
@danielbreves merge? framework is in production.

adammw:
:+1:

PR #23, opened 2016-07-05
Single requests error throw

ocke:
:v:

/cc @zendesk/vegemite

### Description
SDK will throw errors when doing single requests, but not in bulk requests.
Also we now have an errors object which the user can use to look at the errors.

### References
* JIRA: 
* PR: https://github.com/zendesk/zendesk_app_framework/pull/588

### Risks

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -126,6 +126,20 @@ function messageHandler(client, event) {
   }
 }
 
+// When doing singular operations and we retrieve and error this function will throw that error.
+function throwUnlessBulk(path, result) {
+  var isSingularOperation = typeof path === 'string';
+
+  if(isSingularOperation && result.errors[path]) {
+    var err = new Error(result.errors[path].message);
+    err.name = result.errors[path].name;
+    err.stack = result.errors[path].stack;
+    throw err;
+  }
+
+  return result;
+}
+
 var Client = function(options) {
   this._parent = options.parent;
   this._origin = options.origin || this._parent && this._parent._origin;
@@ -428,7 +442,7 @@ Client.prototype = {
       throw new Error('The get method accepts a string or array of strings.');
     }
 
-    return wrappedPostMessage.call(this, 'get', paths);
+    return wrappedPostMessage.call(this, 'get', paths).then(throwUnlessBulk.bind(null, path));
   },
 
   set: function(key, val) {
@@ -446,7 +460,7 @@ Client.prototype = {
       throw new Error('The set method accepts 2 string, or an object.');
     }
 
-    return wrappedPostMessage.call(this, 'set', obj);
+    return wrappedPostMessage.call(this, 'set', obj).then(throwUnlessBulk.bind(null, key));
   },
 
   invoke: function(key) {
@@ -459,7 +473,7 @@ Client.prototype = {
       throw new Error('Invoke with an object isn\'t supported.');
     }
 
-    return wrappedPostMessage.call(this, 'invoke', obj);
+    return wrappedPostMessage.call(this, 'invoke', obj).then(throwUnlessBulk.bind(null, key));
   }
 };
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -72,14 +72,6 @@ function createError(error) {
   return err;
 }
 
-// params is an array for get, and an object for set/invoke
-function createError(error) {
-  var err = new Error(error.msg);
-  err.name = error.code;
-  err.stack = error.stack;
-  return err;
-}
-
 function removePromise(id, timeoutId, args) {
   clearTimeout(timeoutId);
   delete pendingPromises[id];
--- a/lib/client.js
+++ b/lib/client.js
@@ -130,7 +130,8 @@ function messageHandler(client, event) {
 function throwUnlessBulk(path, result) {
   var isSingularOperation = typeof path === 'string';
 
-  if(isSingularOperation && result.errors[path]) {
+  // CRUFT: When framework starts always appending errors we can remove the extra results.errors check
+  if(isSingularOperation && result.errors && result.errors[path]) {
     var err = new Error(result.errors[path].message);
     err.name = result.errors[path].name;
     err.stack = result.errors[path].stack;
--- a/lib/client.js
+++ b/lib/client.js
@@ -64,6 +64,14 @@ function wrappedPostMessage(name, params) {
   return promise.then(removePromise.bind(null, id, timeoutId), removePromise.bind(null, id, timeoutId));
 }
 
+// params is an array for get, and an object for set/invoke
+function createError(error) {
+  var err = new Error(error.msg);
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -104,6 +104,16 @@ function messageHandler(client, event) {
   }
 }
 
+function handleResult(shouldThrow, path, result) {
```
adammw: maybe `throwIfError` or something more specific?

lib/client.js
```diff
@@ -104,6 +104,21 @@ function messageHandler(client, event) {
   }
 }
 
+// When doing singular operations and we retrieve and error this function will throw that error.
+function throwUnlessBulk(path, result) {
+  var isSingularOperation = typeof path === 'string';
+
+  // CRUFT: When framework starts always appending errors we can remove the extra results.errors check
+  if(isSingularOperation && result.errors && result.errors[path]) {
+    var err = new Error(result.errors[path].message);
```
adammw: could we abstract out this since it's used here too: https://github.com/zendesk/zendesk_app_framework_sdk/blob/97db3f093a52b7a79be73918916a0b30511de2c0/lib/client.js#L76-78

lib/client.js
```diff
@@ -374,7 +389,7 @@ Client.prototype = {
       throw new Error('Invoke with an object isn\'t supported.');
     }
 
-    return wrappedPostMessage.call(this, 'invoke', obj);
+    return wrappedPostMessage.call(this, 'invoke', obj).then(throwUnlessBulk.bind(null, key));
```
adammw: Invoke never has bulk operations does it? Can't we continue to use the existing .error property?

lib/client.js
ocke: i was hoping to get rid of that code and change it to errors instead of error

lib/client.js
ocke: not yet, but i was hoping to add that in the future

lib/client.js
```diff
@@ -104,6 +110,19 @@ function messageHandler(client, event) {
   }
 }
 
+// When doing singular operations and we retrieve and error this function will throw that error.
+function throwUnlessBulk(path, result) {
+  var isSingularOperation = typeof path === 'string';
+
+  // CRUFT: When framework starts always appending errors we can remove the extra results.errors check
```
danielbreves: @ocke do you have a matching PR for this?

lib/client.js
```diff
@@ -104,6 +110,19 @@ function messageHandler(client, event) {
...
+  if(isSingularOperation && result.errors) {
+    var err = result.errors.find(function(error) { return error.path === path; });
```
adammw: Array.prototype.find not supported in all browsers...
Also I think this should be an object while results is still an object...

lib/client.js
```diff
@@ -343,7 +359,7 @@ Client.prototype = {
       throw new Error('The get method accepts a string or array of strings.');
     }
 
-    return wrappedPostMessage.call(this, 'get', paths);
+    return wrappedPostMessage.call(this, 'get', paths).then(throwUnlessBulk.bind(null, path));
```
danielbreves: Would be nice if single path `get` returned the result directly, rather than within an object. Though we'd have to make a minor release and document as a breaking change.

lib/client.js
ocke: yes https://github.com/zendesk/zendesk_app_framework/pull/588

lib/client.js
ocke: well, in the current env. yes, but if we are going to change behaviour, i would change the returned result completely and make it more consistent.

lib/client.js
```diff
@@ -126,6 +129,23 @@ function messageHandler(client, event) {
   }
 }
 
+// When doing singular operations and we retrieve and error this function will throw that error.
+function processSingleRequest(path, result) {
```
danielbreves: This seems to process the response, not the request and it's not processing single requests only, can we rename it to `processResponse`?

danielbreves: Can we add specs for single vs bulk request?

spec/client_spec.js
```diff
@@ -275,22 +275,50 @@ describe('Client', function() {
...
+        expect(promise).to.eventually.become({ 'ticket.subject': 'test', 'ticket.requester': 'test' }).and.notify(done);
+
+        window.addEventListener.callArgWith(1, {
+          origin: subject._origin,
+          source: subject._source,
+          data: { id: requestsCount, result: { 'ticket.subject': 'test', 'ticket.requester': 'test' } }
+        });
+      });
+
+      it('doesn\'t accepts multiple arguments', function() {
```
adammw: nit: double quotes over single + escaping?

spec/client_spec.js
```diff
@@ -275,22 +275,50 @@ describe('Client', function() {
...
+        requestsCount--;
```
adammw: this requestsCount is pretty hacky - but we can solve that later in a different PR

spec/client_spec.js
```diff
@@ -309,7 +337,31 @@ describe('Client', function() {
...
+
+      it('throws an error, which you can catch, when requesting something that doesn\'t exist', function(done) {
+        var promise = subject.get('ticket.subj');
+
+        expect(promise).to.be.rejectedWith(Error, 'No such Api').and.notify(done);
+
+        window.addEventListener.callArgWith(1, {
+          origin: subject._origin,
+          source: subject._source,
+          data: { id: requestsCount, result: { errors: { 'ticket.subj': {message: 'No such Api'} } } }
```
adammw: lint: spaces around message property in object.

spec/client_spec.js
```diff
@@ -309,7 +337,31 @@ describe('Client', function() {
...
+        });
+      });
+
+      it('does not throw an error when bulk requesting', function(done) {
+        var promise = subject.get(['ticket.subj']);
+
+        expect(promise).to.become({ errors: {'ticket.subj': { message: 'No such Api'} } }).and.notify(done);
```
adammw: lint: spaces around 'ticket.subj'

spec/client_spec.js
```diff
@@ -275,22 +275,50 @@ describe('Client', function() {
         requestsCount++;
       });
 
-      it('returns a promise', function() {
+      it('returns a promise', function(done) {
         promise = subject.get('ticket.subject');
 
         expect(promise).to.be.a.promise;
+        expect(promise).to.eventually.become('test').and.notify(done);
```
adammw: protip: if you don't want to have to notify done all the time, you can just return the promise to chai and it'll handle it.

ocke: oh, you should've told me that earlier.

danielbreves:
👍  though this will require a minor bump.

sandlerr:
:+1:

PR #24, opened 2016-07-11
[BUG] Fixes an issue where Save Hook rules aren't being respect for synchronous return

svizzari:
:v:

/cc @zendesk/vegemite

### Description

Where a Save Hook handler was returning immediately, rather than a Promise, the rules governing Save Hook return values weren't being respected.

### TODO
- [x] ~~Add regression test~~ 04a8150 59a7ac7

### Risks
* [low] Save Hook allowing Ticket Save when not intended, or vice-versa

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -88,9 +88,7 @@ function when(items) {
         } else if (isFalsy(res)) {
           promise = Promise.reject(res);
         } else {
-          promise = new Promise(function(resolve) {
-            resolve(res);
-          });
+          promise = Promise.resolve(res);
         }
       } catch(err) {
         promise = Promise.reject(err);
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -85,15 +85,15 @@ function when(items) {
         res = item();
         if (isPromise(res)) {
           promise = res;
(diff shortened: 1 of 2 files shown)


danielbreves:
👍  on 💚

lib/utils.js
```diff
@@ -85,15 +85,15 @@ function when(items) {
         res = item();
         if (isPromise(res)) {
           promise = res;
+        } else if (isFalsy(res)) {
+          promise = Promise.reject(res);
         } else {
           promise = new Promise(function(resolve) {
```
adammw: Promise.resolve, if i want to be a stickler

adammw:
:+1:

lib/utils.js
```diff
@@ -85,15 +85,13 @@ function when(items) {
         res = item();
         if (isPromise(res)) {
           promise = res;
+        } else if (isFalsy(res)) {
+          promise = Promise.reject(res);
         } else {
-          promise = new Promise(function(resolve) {
-            resolve(res);
-          });
+          promise = Promise.resolve(res);
```
svizzari: @adammw 💇 To make "the stickler" less sticky

PR #25, opened 2016-07-12
[AF-359] Add app instances API for v2

adammw:
:v: 🚧 

/cc @zendesk/vegemite

### Description
See https://github.com/zendesk/zendesk_app_framework/pull/589

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-359

### Risks
- ?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -180,6 +180,7 @@ Client.prototype = {
   /// client.postMessage('hello', { awesome: true });
   /// ```
   postMessage: function(name, data) {
+    if (this._parent) { throw new Error('Not Implemented'); }
     var msg = JSON.stringify({ key: name, message: data, appGuid: this._appGuid, instanceGuid: this._instanceGuid });
     rawPostMessage(this, msg, name === 'iframe.handshake');
   },
@@ -210,6 +211,7 @@ Client.prototype = {
   /// `app.registered` event. You can add as many handlers to `app.registered` as you like. They're
   /// called immediately after the `init` callback.
   on: function(name, handler, context) {
+    if (this._parent) { throw new Error('Not Implemented'); }
     if (typeof handler == 'function') {
       handler = context ?
         handler.bind(context) :
@@ -242,6 +244,7 @@ Client.prototype = {
   /// });
   /// ```
   off: function(name, handler) {
+    if (this._parent) { throw new Error('Not Implemented'); }
     if (!this._messageHandlers[name]) { return false; }
     var index = this._messageHandlers[name].indexOf(handler);
     this.postMessage('iframe.off:' + name);
@@ -268,6 +271,7 @@ Client.prototype = {
   /// client.has('app.activated', appRegistered); // false
   /// ```
   has: function(name, handler) {
+    if (this._parent) { throw new Error('Not Implemented'); }
     if (!this._messageHandlers[name]) { return false; }
     return this._messageHandlers[name].indexOf(handler) !== -1;
   },
@@ -291,6 +295,7 @@ Client.prototype = {
   /// client.trigger('activation') // activating!
   /// ```
   trigger: function(name, data) {
+    if (this._parent) { throw new Error('Not Implemented'); }
     if (!this._messageHandlers[name]) { return false; }
     this._messageHandlers[name].forEach(function(handler) {
       handler(data);
@@ -322,6 +327,7 @@ Client.prototype = {
   /// );
   /// ```
   request: function(options) {
+    if (this._parent) { return this._parent.request(options); }
     var requestKey = 'request:' + nextIdFor('request');
 
     return new Promise(function(resolve, reject) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -401,7 +401,7 @@ Client.prototype = {
   },
 
   context: function() {
-    if (this._instanceGuid) {
+    if (this._instanceGuid && this._instanceGuid != this._appGuid) {
       var key = 'instances.' + this._instanceGuid;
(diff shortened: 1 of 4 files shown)


lib/client.js
```diff
@@ -311,7 +341,18 @@ Client.prototype = {
     }.bind(this));
   },
 
+  instance: function(instanceGuid) {
```
svizzari: Could have avoided each sub client having the `instanceClients` and this guard by making `instances` a static method, like `init`?

lib/client.js
```diff
@@ -104,16 +126,24 @@ function messageHandler(client, event) {
   }
 }
 
-var Client = function(origin, appGuid) {
-  this._origin = origin;
+var Client = function(options) {
+  this._parent = options.parent;
```
svizzari: I'd think I'd prefer `root` to `parent`, which seems loaded with meaning in iframe context

svizzari:
Very neat! I like this, it's kinda how I thought the bare client could be used directly. Let's get tests passing tomorrow 😹

lib/client.js
```diff
@@ -311,7 +341,48 @@ Client.prototype = {
...
+  ///
+  /// ##### Returns
+  ///
+  /// A [Promises/A+](https://promisesaplus.com) conformant `promise` object.
+  ///
+  /// ```javascript
+  /// var client = ZAFClient.init();
+  ///
+  /// client.get('instances').then(function(data) {
+  ///   var guid = Object.keys(data.instances).filter(function(guid) {
```
ocke: probably because I just hada discussion on APIs and associative arrays, but I think objects as arrays is fine for internal mechanism, but terrible for public APIs.

ocke:
I like your code cleanup with add/removing spaces/newlines.

lib/client.js
adammw: what about the data hanging off objects? isn't that the general idea of a data-node like structure? I mean, the alternative is having an array that you have to look inside each and grab about the `instanceGuid` property...

svizzari:
👍

lib/client.js
```diff
@@ -311,7 +347,48 @@ Client.prototype = {
...
+  /// }).then(function(guid) {
+  ///   return client.instance(guid);
+  /// }).then(function(topBarClient) {
+  ///   topBarClient.invoke('popover');
+  /// });
+  /// ```
+  instance: function(instanceGuid) {
+    if (this._parent) { return this._parent.instance(instanceGuid); }
+    var instanceContextKey = 'instances.' + instanceGuid;
+    this._instanceClients[instanceGuid] = this._instanceClients[instanceGuid] ||
```
sandlerr: any particular reason it needs to be stored on `this` rather than just a private var in this file?

sandlerr: would we ever be caching rejected promises?

adammw: I guess in the structure of it, the instance clients have a parent, so you wouldn't want the wrong child going to the wrong parent?

sandlerr: fair enough, we'd need an extra level of nesting in the shared object just to keep it private. Carry on

sandlerr:
:+1:

lib/client.js
adammw: rejected promises should bubble up back to the user with the rejection I would have thought.
I guess the only problem there is that it couldn't retry - but it seems like such an edge case it's not worth worrying about.

lib/client.js
```diff
@@ -324,6 +401,14 @@ Client.prototype = {
   },
 
   context: function() {
+    if (this._instanceGuid) {
```
adammw: whoops... this line will break context for existing apps... pushing a commit to fix

lib/client.js
```diff
@@ -311,7 +347,48 @@ Client.prototype = {
...
+  /// ```javascript
+  /// var client = ZAFClient.init();
+  ///
+  /// client.get('instances').then(function(data) {
+  ///   var guid = Object.keys(data.instances).filter(function(guid) {
+  ///     return data.instances[guid].location == 'top_bar';
+  ///   })[0];
+  ///   return guid;
+  /// }).then(function(guid) {
+  ///   return client.instance(guid);
```
danielbreves: Would be nice if this returned a client rather than a promise, there's a fair amount of boilerplate already.

danielbreves: So it should be possible to do something like this:
```js
client.get('instances').then(function(data) {
  var guid = Object.keys(data.instances).filter(function(guid) {
    return data.instances[guid].location == 'top_bar';
  })[0];

  client.instance(guid).invoke('popover');
});
```

sandlerr: what use is the guid other than to construct clients? Why shouldn't `client.get('instances')` resolve with an array/object of clients?

adammw: get only deals with piping stuff in and out of the message pipe, so it only really deals with JSON-serializable objects. i wouldn't feel comfortable special casing certain request paths like that...

danielbreves: Makes sense, though `instance` is a new API and it doesn't need any extra information from ZAF to construct a client, does it?

adammw: > So it should be possible to do something like this:

my question is that how, in that scenario, would we let the developer know they specified a wrong guid or an app they don't have access too? I guess it'd be okay, but we're tying us into being synchronous always.

danielbreves: Couldn't we validate the `guid` when `get`, `set` or `invoke` is called. That's an unlikely scenario in any case, isn't it? Also, we'd only be locking us into being synchronous on this minor release, right?

PR #27, opened 2016-07-25
Add subscriber count to iframe event handler events

adammw:
:v:

/cc @zendesk/vegemite

### Description
Add subscriber count to iframe event handler events.
This allows the framework to know how many subscribers are listening inside the iframe, and when it's safe to bind/unbind.

### References
* JIRA: 

### Risks
* low: data attribute of event is unused

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -248,7 +248,9 @@ Client.prototype = {
     if (this._parent) { throw new Error('Not Implemented'); }
     if (!this._messageHandlers[name]) { return false; }
     var index = this._messageHandlers[name].indexOf(handler);
-    this._messageHandlers[name].splice(index, 1);
+    if (this.has(name, handler)) {
+      this._messageHandlers[name].splice(index, 1);
+    }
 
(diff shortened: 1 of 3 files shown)


danielbreves:
👍

svizzari:
👍

sandlerr:
@zendesk/vegemite I just pushed one more commit, can we have one more thumb please?

svizzari:
>  I just pushed one more commit, can we have one more thumb please?

oh, yea, that'd be bad if we started splicing elements from the end of the array when start index was negative

👍

lib/client.js
```diff
@@ -248,7 +248,9 @@ Client.prototype = {
     if (this._parent) { throw new Error('Not Implemented'); }
     if (!this._messageHandlers[name]) { return false; }
     var index = this._messageHandlers[name].indexOf(handler);
-    this._messageHandlers[name].splice(index, 1);
+    if (this.has(name, handler)) {
```
adammw: wouldn't `index !== -1` be equivilant? #codegolf

PR #30, opened 2016-08-04
Save hook should resolve when return is omitted

danielbreves:
:v:

/cc @zendesk/vegemite

So it's consistent with v1: https://github.com/zendesk/zendesk_app_framework/blob/master/lib/events/hook.js#L43

### Risks
* [low] Could break v2 save hook

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -33,7 +33,7 @@ function isPromise(obj) {
     !!obj && obj.then && typeof obj.then === 'function';
 }
 
-function isFalsy(val) {
+function shouldReject(val) {
   return val === false ||
     val instanceof Error ||
     typeof val === 'string';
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -34,7 +34,7 @@ function isPromise(obj) {
 }
 
 function isFalsy(val) {
-  return !val ||
+  return val === false ||
```
svizzari: Pretty sure you asked me to change this 😿

sandlerr:
:+1:

lib/utils.js
```diff
@@ -34,7 +34,7 @@ function isPromise(obj) {
 }
 
 function isFalsy(val) {
```
ocke: The function name 'isFalsy' isn't correct.

svizzari: `isFalsy('The previous comment ^') === true`

svizzari:
:+1:

lib/utils.js
danielbreves: I think my comment was about `!val || val === false` being redundant. But I didn't realise `!val` would cause handlers without a `return` statement to block the save hook, which doesn't happen in v1.

danielbreves:
@zendesk/vegemite 💇

sandlerr:
LGTM

PR #32, opened 2016-08-05
Return full results, and tests

ocke:
:v:

/cc @zendesk/vegemite

### Description
Added tests and tests and tests.
Reverted 1 breaking change: always return full result

### References
* JIRA: 

### Risks

### DIFF
--- a/spec/client_spec.js
+++ b/spec/client_spec.js
@@ -299,12 +299,9 @@ describe('Client', function() {
     });
 
     describe('#get', function() {
-
-
-      it('returns a promise', function(done) {
+      it('takes an argument and returns a promise with data', function(done) {
         promise = subject.get('ticket.subject');
 
-        expect(promise).to.be.a.promise;
         expect(promise).to.eventually.become({ errors: {}, 'ticket.subject': 'test' }).and.notify(done);
 
         window.addEventListener.callArgWith(1, {
@@ -317,7 +314,6 @@ describe('Client', function() {
       it('throws an error when the handler throws it', function(done) {
         promise = subject.get('ticket.err');
 
-        expect(promise).to.be.a.promise;
         expect(promise).to.be.rejectedWith(Error, 'ticket.err unavailable').and.notify(done);
 
         window.addEventListener.callArgWith(1, {
@@ -330,7 +326,6 @@ describe('Client', function() {
       it('accepts an array with multiple paths', function(done) {
         promise = subject.get(['ticket.subject', 'ticket.requester']);
 
-        expect(promise).to.be.a.promise;
         expect(promise).to.eventually.become({
           'ticket.subject': 'test',
           'ticket.requester': 'test'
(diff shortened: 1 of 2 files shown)


spec/client_spec.js
```diff
@@ -333,12 +331,30 @@ describe('Client', function() {
...
+          origin: subject._origin,
+          source: subject._source,
+          data: { id: requestsCount, result: {
+            'ticket.subject': 'test',
+            'ticket.requester': 'test'
+          }}
+        });
+      });
+
+      it('succeeds with errors when bulk requesting', function(done) {
```
sandlerr: maybe `resolves` instead of `succeeds`?

spec/client_spec.js
```diff
@@ -357,73 +373,102 @@ describe('Client', function() {
...
-      it('does not throw an error when bulk requesting', function(done) {
-        var promise = subject.get(['ticket.subj']);
+      it('accepts an object with multiple paths', function(done) {
+        promise = subject.set({
+          'ticket.subject': 'value',
+          'ticket.description': 'value'
+        });
 
-        expect(promise).to.become({ errors: { 'ticket.subj': { message: 'No such Api' } } }).and.notify(done);
+        expect(promise).to.be.a.promise;
```
sandlerr: isn't this implied by the next line? or is the error message really bad when it fails?

spec/client_spec.js
```diff
@@ -435,19 +480,48 @@ describe('Client', function() {
     });
 
     describe('#invoke', function() {
-      it('returns a promise', function() {
-        promise = subject.invoke('iframe.resize');
+      it('returns a promise', function(done) {
```
sandlerr: this description doesn't match what the test actually does

spec/client_spec.js
```diff
@@ -435,19 +480,48 @@ describe('Client', function() {
...
+
+        window.addEventListener.callArgWith(1, {
+          origin: subject._origin,
+          source: subject._source,
+          data: { id: requestsCount, result: { errors: {}, 'ticket.appendText': true } }
+        });
       });
 
-      it('throw when called with an object', function() {
+      it('throws an error when the handler throws it', function(done) {
```
sandlerr: the handler isn't exactly throwing an error, it's just returning one

spec/client_spec.js
```diff
@@ -435,19 +480,48 @@ describe('Client', function() {
...
+          data: { id: requestsCount, result: { errors: { 'ticket.foo': { message: 'ticket.foo unavailable' } } } }
+        });
+      });
+
+      it('throws an error when invoked with an object', function() {
         expect(function() {
           subject.invoke({
             'iframe.resize': [1]
           });
         }).to.throw(Error);
```
sandlerr: while you're here, can we make this assertion more specific? I'd like to make sure that this error has a useful message for developers.

sandlerr:
I don't really like putting `expect(promise).to.be.a.promise;` everywhere. Seems like something that could maybe be provided by `chai-as-promised`?

spec/client_spec.js
ocke: It has a good messge, it just that this test stubs it all. in reality the message would be the error message from console.

ocke:
You're right, I removed them

spec/client_spec.js
sandlerr: nvm, I am wrong

sandlerr:
:+1: on green

danielbreves:
👍

PR #26, opened 2016-07-20
Allow event handlers in clients for other instances

adammw:
:v:

/cc @zendesk/vegemite

**depends on changes to App Framework**

### Description
Allow event handlers / trigger in clients for other instances

### Tasks
- [ ] Get +1 of team
- [ ] Create changes to App Framework to proxy events between instances
- [ ] Make trigger fire events in other instances as well as just local handlers

### Risks
* low: new feature

PR #34, opened 2016-08-18
[AF-359] Add support for binding and triggering events across instances

danielbreves:
:v:

✅  _Depends on https://github.com/zendesk/zendesk_app_framework/pull/622_

/cc @zendesk/vegemite

### Description
Add support for binding and triggering events across instances. This PR also changes the `instance` API to return a client rather than a promise that resolves with a client. See discussion [here](https://github.com/zendesk/zendesk_app_framework_sdk/pull/25/files#r71476024).

### Tasks
- [x] Write tests

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-359

### Risks
* [low] instances API may not work as intended.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -90,7 +90,7 @@ function triggerEvent(client, eventName, data) {
   });
 }
 
-function finalisePendingPromise(pendingPromise, data) {
+function finalizePendingPromise(pendingPromise, data) {
   if (data.error) {
     var error = data.error.name ? createError(data.error) : data.error;
     pendingPromise.reject(error);
@@ -142,7 +142,7 @@ function messageHandler(client, event) {
       pendingPromise;
 
   if (data.id && (pendingPromise = pendingPromises[data.id])) {
-    finalisePendingPromise(pendingPromise, data);
+    finalizePendingPromise(pendingPromise, data);
   } else if (ZAF_EVENT.test(data.key)) {
     var key = data.key.replace(ZAF_EVENT, ''),
         msg = { appGuid: client._appGuid };
--- a/lib/client.js
+++ b/lib/client.js
@@ -391,6 +391,7 @@ Client.prototype = {
   },
 
   instance: function(instanceGuid) {
+    if (instanceGuid === this._instanceGuid) { return this; }
     if (this._parent) { return this._parent.instance(instanceGuid); }
     var instanceClient = this._instanceClients[instanceGuid];
     if (!instanceClient) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -98,6 +98,8 @@ function messageHandler(client, event) {
     }
   }
 
+  if (data.instanceGuid && data.instanceGuid !== client._instanceGuid) { return; }
+
   var pendingPromise;
   if (data.id && (pendingPromise = pendingPromises[data.id])) {
     if (data.error) {
@@ -124,7 +126,10 @@ function messageHandler(client, event) {
         delete client._repliesPending[key];
       });
     } else {
-      client.trigger(key, data.message);
+      if (!client._messageHandlers[key]) { return; }
+      client._messageHandlers[key].forEach(function(handler) {
+        handler(data.message);
+      });
     }
   }
 }
@@ -159,11 +164,6 @@ var Client = function(options) {
   this._context = options.context || null;
   this.ready = false;
 
-  if (this._parent) {
-    this.ready = this._parent.ready;
-    return this; // shortcut handshake
-  }
-
   this.on('app.registered', function(data) {
     this.ready = true;
     this._metadata = data.metadata;
@@ -174,9 +174,14 @@ var Client = function(options) {
     this._context = context;
   }, this);
 
-  this.postMessage('iframe.handshake', { version: version });
-
   window.addEventListener('message', messageHandler.bind(null, this));
+
+  if (this._parent) {
+    this.ready = this._parent.ready;
+    return this; // shortcut handshake
+  }
+
+  this.postMessage('iframe.handshake', { version: version });
 };
 
 Client.prototype = {
@@ -200,7 +205,6 @@ Client.prototype = {
   /// client.postMessage('hello', { awesome: true });
   /// ```
   postMessage: function(name, data) {
-    if (this._parent) { throw new Error('Not Implemented'); }
     var msg = JSON.stringify({ key: name, message: data, appGuid: this._appGuid, instanceGuid: this._instanceGuid });
     rawPostMessage(this, msg, name === 'iframe.handshake');
   },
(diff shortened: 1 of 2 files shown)


danielbreves:
I wonder if we should limit the access for child instances to `get`, and `on`.

svizzari:
👍

lib/client.js
```diff
@@ -159,11 +164,6 @@ var Client = function(options) {
   this._context = options.context || null;
   this.ready = false;
 
-  if (this._parent) {
-    this.ready = this._parent.ready;
-    return this; // shortcut handshake
-  }
-
   this.on('app.registered', function(data) {
```
adammw: so we want to listen to these events now for other instances? are they ever fired in time?
my original thinking was that the metadata should be shared with the parent (since it doesn't change between locations), and the context could be retrieved from the instances API.

adammw:
:cry: 💔  https://github.com/zendesk/zendesk_app_framework_sdk/pull/26

there's a couple of things that i think should be ported across, like the test fixes and extracting the `triggerEvent`

lib/client.js
danielbreves: Yea, I think it should fire in time if your app itself is creating the instance (i.e. modals). But if it doesn't it shouldn't matter, the parent app can still retrieve the context using the `instances` or `context` API from the instance client.

lib/client.js
```diff
@@ -83,6 +83,42 @@ function isValidEvent(client, event) {
...
 }
 
+function triggerEvent(client, eventName, data) {
+  if (!client._messageHandlers[eventName]) { return false; }
+  client._messageHandlers[eventName].forEach(function(handler) {
+    handler(data);
+  });
+}
+
+function isIntendedRecipient(recipientGuid, clientGuid) {
```
svizzari: Would `isAnotherInstance` be a more appropriate name?

lib/client.js
```diff
@@ -83,6 +83,42 @@ function isValidEvent(client, event) {
...
+  return recipientGuid && recipientGuid !== clientGuid;
+}
+
+function handlePendingPromise(pendingPromise, data) {
```
svizzari: `finalisePendingPromise` or `commitPendingPromise`?

lib/client.js
```diff
@@ -321,11 +339,7 @@ Client.prototype = {
   /// client.trigger('activation') // activating!
   /// ```
   trigger: function(name, data) {
-    if (this._parent) { throw new Error('Not Implemented'); }
-    if (!this._messageHandlers[name]) { return false; }
-    this._messageHandlers[name].forEach(function(handler) {
-      handler(data);
-    });
+    this.postMessage('iframe.trigger:' + name, { message: data });
```
adammw: do we not want to call `triggerEvent` as well? what about triggering an event in the same instance? (do we have specs for that?)

lib/client.js
```diff
@@ -83,6 +83,42 @@ function isValidEvent(client, event) {
...
+function handlePendingPromise(pendingPromise, data) {
+  if (data.error) {
+    var error = data.error.name ? createError(data.error) : data.error;
+    pendingPromise.reject(error);
+  } else {
+    pendingPromise.resolve(data.result);
+  }
+}
+
+function postReply(client, key, msg) {
```
svizzari: `postReplyWith`?

😅

lib/client.js
```diff
@@ -375,16 +389,15 @@ Client.prototype = {
...
-    return this._instanceClients[instanceGuid];
+    var instanceClient = this._instanceClients[instanceGuid];
+    if (!instanceClient) {
+      instanceClient = new Client({
+        parent: this,
+        instanceGuid: instanceGuid
+      });
+      this._instanceClients[instanceGuid] = instanceClient;
+    }
+    return instanceClient;
```
adammw: I assume instance API never went public?

spec/client_spec.js
```diff
@@ -542,38 +556,31 @@ describe('Client', function() {
       });
 
       it('requests instance context and caches it in the returned client', function() {
```
adammw: rename spec now behaviour has changed

lib/client.js
danielbreves: Triggering an event on the same instance works through the framework as well. We could potentially short-circuit it so we manually call the handlers here and ignore incoming events from the framework if they were triggered by the instance, but right now there's no way to tell when events are triggered by the instance.

lib/client.js
```diff
@@ -149,7 +175,7 @@ function processResponse(path, result) {
 var Client = function(options) {
   this._parent = options.parent;
   this._origin = options.origin || this._parent && this._parent._origin;
-  this._source = window.parent;
+  this._source = options.source || this._parent && this._parent._source || window.parent;
```
sandlerr: could extract `fetchFromParentWithDefault()`

lib/client.js
sandlerr: Using the instances API to fetch an instance, and then using the instances API from that instance?

lib/client.js
sandlerr: this could be a breaking change right? synchronous to async?

danielbreves: Mmm good point.

lib/client.js
```diff
@@ -98,33 +138,19 @@ function messageHandler(client, event) {
...
+      pendingPromise;
+
   if (data.id && (pendingPromise = pendingPromises[data.id])) {
-    if (data.error) {
-      var error = data.error.name ? createError(data.error) : data.error;
-      pendingPromise.reject(error);
-    } else {
-      pendingPromise.resolve(data.result);
-    }
+    finalisePendingPromise(pendingPromise, data);
```
sandlerr: nit: we tend to use American spelling

lib/client.js
danielbreves: Like this? I doesn't feel like much of an improvement...
```
function fetchFromParentWithDefault(parent, prop, defaultValue) {
  return this._parent && this._parent[prop] || defaultValue;
}
// ...
  this._origin = options.origin || fetchFromParentWithDefault(this._parent, '_origin');
  this._source = options.source || fetchFromParentWithDefault(this._parent, '_source', window.parent);
  this._appGuid = options.appGuid || fetchFromParentWithDefault(this._parent, '_appGuid');
```

sandlerr:
:+1:

lib/client.js
```diff
@@ -83,6 +83,46 @@ function isValidEvent(client, event) {
...
+    };
+    rawPostMessage(client, msg);
+  }).then(function() {
+    delete client._repliesPending[key];
+  });
+}
+
+function getMessageRecipient(client, recipientGuid) {
+  if (recipientGuid && recipientGuid !== client._instanceGuid) {
+    return client._instanceClients[recipientGuid];
```
adammw: it's possible that this returns `null` right? how would we handle that?

adammw:
:+1:

lib/client.js
danielbreves: Mmm I don't see how, because if the framework knows posts a message to an iframe with an `instanceGuid` it's because this iframe create a client for that instance at some point, and we're never clearing `_instanceClients` right now. But I agree we should still handle that. I'll make it log an error.

danielbreves:
@adammw 💇

adammw:
:+1:

PR #36, opened 2016-09-08
AF-293 whitelist slow requests from being auto-rejected, remove JSHint

sandlerr:
cc @zendesk/vegemite 

Allows some requests not to get auto-rejected since we anticipate they can take over 5 seconds in some non-error cases. Required for opening modals.

### Risks
- [medium] the sdk might crash if it fails to process empty values as the arguments to a get, set, or invoke call.


lib/client.js
```diff
@@ -1,11 +1,38 @@
...
+  case 'invoke': {
+    var allowed = ['instances.create'];
+    if (allowed.length !== paramsArray.length) {
+      break;
+    }
+    var match = paramsArray.every(function(value, index) {
+      return value === allowed[index];
+    });
+    if (match) {
+      return NaN; // timer IDs are numbers; embrace the JavaScript 😬
```
svizzari: bwahaha, embrace the js

svizzari:
I _guess_ one alternative might have been to use the whitelist to allow retries on the existing timeout, rather than indeterminate delay.  It'd result in _n_ multiples of 5 secs, let's say three for the sake of argument.

sandlerr:
There's no limit on how long a previous modal might have been left open. If we reject this promise then the caller never gets notified about the guid of the new instance. If we need another API where we can be almost certain that an error has occurred after 15 or 20 seconds with no response, we should implement retries or a longer timeout then. I don't think that's the case for this API.

lib/client.js
```diff
@@ -1,11 +1,38 @@
...
 
+// @internal
+// #timeoutReject(name, params)
+//
+// Reject a request if required, given the rejection function, name (get, set or invoke)
+// and arguments to that request. Falls back to 5000 second timeout with few exceptions.
+//
+function timeoutReject(reject, name, paramsArray) {
+  switch(name) {
+  case 'invoke': {
```
danielbreves: Shouldn't this be indented?

spec/client_spec.js
```diff
@@ -574,6 +575,19 @@ describe('Client', function() {
         clock.restore();
         expect(promise).to.be.rejectedWith(Error, 'Invocation request timeout').and.notify(done);
       });
+
+      it('doesnt reject whitelisted promises after 5 seconds', function(done) {
+        var clock = sinon.useFakeTimers();
+        promise = subject.invoke('instances.create');
+        clock.tick(10000);
+        clock.restore();
+        expect(promise).to.eventually.become({ errors: {}, 'ticket.appendText': true }).and.notify(done);
```
danielbreves: `ticket.appendText` is confusing here, can we use the name of the request 'instances.create'?

lib/client.js
```diff
@@ -1,11 +1,38 @@
...
+    }
+    var match = paramsArray.every(function(value, index) {
+      return value === allowed[index];
+    });
+    if (match) {
+      return NaN; // timer IDs are numbers; embrace the JavaScript 😬
+    }
+  }
+  }
+  return setTimeout(function() {
```
danielbreves: So if you make a bulk `invoke` (not supported at the moment) with both allowed and not allowed APIs it can timeout?

lib/client.js
sandlerr: Not according to the current eslint config

sandlerr:
@danielbreves 💇 ?

lib/client.js
danielbreves: Can we change the config? I find it pretty weird to have the two closing brackets on the same indentation level below. 😛

lib/client.js
```diff
@@ -1,11 +1,38 @@
...
+// @internal
+// #timeoutReject(name, params)
+//
+// Reject a request if required, given the rejection function, name (get, set or invoke)
+// and arguments to that request. Falls back to 5000 second timeout with few exceptions.
+//
+function timeoutReject(reject, name, paramsArray) {
+  switch(name) {
+  case 'invoke': {
+    var allowed = ['instances.create'];
```
danielbreves: Can we make this a constant at the top of the file?

lib/client.js
```diff
@@ -1,11 +1,38 @@
...
+    if (allowed.length !== paramsArray.length) {
+      break;
+    }
+    var match = paramsArray.every(function(value, index) {
+      return value === allowed[index];
```
danielbreves: Why do the indexes have to match?

sandlerr: this was comparing two arrays element by element (`allowed` and `paramsArray`)

lib/client.js
sandlerr: done, 💇 ?

sandlerr:
@danielbreves @zendesk/vegemite :+1:?

lib/client.js
```diff
@@ -1,11 +1,45 @@
...
+// @internal
+// #timeoutReject(name, params)
+//
+// Reject a request if required, given the rejection function, name (get, set or invoke)
+// and arguments to that request. Falls back to 5000 second timeout with few exceptions.
+//
+function timeoutReject(reject, name, paramsArray) {
+  switch(name) {
+    case 'invoke': {
+      var matches = paramsArray.map(function(action) {
```
danielbreves: Could you use `filter` here and then 
```
var allWhitelisted = matches.length === paramsArray.length;
var noneWhitelisted = matches.length === 0;
```

spec/client_spec.js
```diff
@@ -574,6 +577,19 @@ describe('Client', function() {
...
+        promise = subject.invoke('instances.create');
+        clock.tick(10000);
+        clock.restore();
+        expect(promise).to.eventually.become({ errors: {}, 'instances.create': { url: 'http://a.b' } }).and.notify(done);
+        window.addEventListener.callArgWith(1, {
+          origin: subject._origin,
+          source: subject._source,
+          data: { id: requestsCount, result: { errors: {}, 'instances.create': { url: 'http://a.b' } } }
+        });
+      });
```
danielbreves: Could we add a test for the "Illegal bulk call" case as well?

sandlerr: Nope, `wrappedPostMessage` is not exported and invoke will throw unless it's called with a single string argument.

lib/client.js
sandlerr: done, 💇 ?

danielbreves:
👍

lib/client.js
```diff
@@ -1,11 +1,41 @@
...
     pendingPromises = {},
     ids             = {};
 
+// @internal
+// #timeoutReject(name, params)
+//
+// Reject a request if required, given the rejection function, name (get, set or invoke)
+// and arguments to that request. Falls back to 5000 second timeout with few exceptions.
+//
+function timeoutReject(reject, name, paramsArray) {
```
svizzari: #nit `timeoutReject` bothers me slightly, like it takes a little more effort to understand what it's actually doing.

svizzari:
👍

lib/client.js
sandlerr: as in the name of the function?

svizzari: Yup, that was what I was referring to

PR #38, opened 2016-10-03
[AF-488] Support hash params for origin and app guid

adammw:
:v:

/cc @zendesk/vegemite

### Description
Support hash params for origin and app guid

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-488

### Risks
* medium: handshake breaks, causing all v2 api calls to fail/error out

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -25,12 +25,15 @@ var Client    = require('client'),
 /// });
 /// ```
 ZAFClient.init = function(callback) {
-  var params = Utils.queryParameters(),
+  var queryParams = Utils.queryParameters(),
+      hashParams = Utils.queryParameters(( document.location.hash || '' ).slice(1)),
+      origin = queryParams.origin || hashParams.origin,
+      app_guid = queryParams.app_guid || hashParams.app_guid,
       client;
 
-  if (!params.origin || !params.app_guid) { return false; }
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -25,12 +25,15 @@ var Client    = require('client'),
 /// });
 /// ```
 ZAFClient.init = function(callback) {
-  var params = Utils.queryParameters(),
+  var queryParams = Utils.queryParameters(),
+      hashParams = Utils.queryParameters(( document.location.hash || '' ).slice(1)),
```
sandlerr: redundant `document`?

adammw: possibly, had copied from the source of `Utils.queryParameters` which did the same thing... doesn't hurt does it?

lib/index.js
```diff
@@ -25,12 +25,15 @@ var Client    = require('client'),
...
+      origin = queryParams.origin || hashParams.origin,
+      app_guid = queryParams.app_guid || hashParams.app_guid,
```
danielbreves: `appGuid`?

adammw: this is the the query parameter, which uses snake case...

danielbreves: I know, we need to leave the parameter as is, but we can rename the `var`, no?

PR #39, opened 2016-10-14
Send caught error responses when replying

sandlerr:
[postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) uses the [structured clone algo](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) which does not like sending error objects.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -138,7 +138,7 @@ function postReplyWith(client, key, msg) {
   if (client._repliesPending[key]) { return; }
   msg.key = 'iframe.reply:' + key;
   client._repliesPending[key] = true;
-  when(client._messageHandlers[key]).then(
+  return when(client._messageHandlers[key]).then(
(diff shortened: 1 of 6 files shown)


lib/client.js
```diff
@@ -131,11 +138,19 @@ function postReplyWith(client, key, msg) {
...
     rawPostMessage.bind(null, client, msg)
   ).catch(function(reason) {
+    var rejectMessage;
+    // if the handler throws an error we want to send back its message, but
+    // Error objects don't pass through the postMessage boundary.
+    if (reason instanceof Error) {
+      rejectMessage = reason.message;
+    } else {
+      rejectMessage = reason;
+    }
```
danielbreves: `var rejectMessage = reason instanceof Error ? reason.message : reason;`?

PR #41, opened 2016-10-28
Throw an Error when an instanceGuid is not passed to instance API

danielbreves:
:v:

/cc @zendesk/vegemite

### Risks
* [low] Could break apps using the instance API (closed beta only)


sandlerr:
why not check `typeof`?

danielbreves:
@sandlerr :haircut_man:

PR #44, opened 2017-05-30
[AF-684] Promises only take 1 argument

ocke:
:v:
/cc @zendesk/vegemite
/cc @zendesk/zdbreaking 

### Description
Resolve and reject only takes 1 argument.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-684

### Risks
* [high] Apps will almost definitely rely on this bug.

### DIFF
--- a/spec/client_spec.js
+++ b/spec/client_spec.js
@@ -419,7 +419,7 @@ describe('Client', function() {
         it('resolves when the request succeeds', function(done) {
           triggerEvent(subject, 'request:' + requestsCount + '.done', response);
           promise.then(function() {
-            expect(doneHandler).to.have.been.calledWith(response.responseArgs[0]);
+            expect(doneHandler).to.have.been.calledWith(response.responseArgs);
             done();
           });
         });
@@ -427,7 +427,7 @@ describe('Client', function() {
         it('rejects when the request fails', function(done) {
           triggerEvent(subject, 'request:' + requestsCount + '.fail', response);
           promise.then(function() {
-            expect(failHandler).to.have.been.calledWith(response.responseArgs[0]);
+            expect(failHandler).to.have.been.calledWith(response.responseArgs);
             done();
           });
         });
--- a/lib/client.js
+++ b/lib/client.js
@@ -423,11 +423,11 @@ Client.prototype = {
       }
 
       this.on(requestKey + '.done', function(evt) {
-        resolve.apply(this, evt.responseArgs);
+        resolve.call(this, evt.responseArgs);
       });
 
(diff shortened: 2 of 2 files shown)


spec/client_spec.js
```diff
@@ -419,15 +419,15 @@ describe('Client', function() {
         it('resolves when the request succeeds', function(done) {
           triggerEvent(subject, 'request:' + requestsCount + '.done', response);
           promise.then(function() {
-            expect(doneHandler).to.have.been.calledWith(response.responseArgs[0]);
```
danielbreves: @ocke so now request handlers will always receive an array of arguments? This will definitely break apps. I think we need a different strategy.

dpawluk: @ocke definitely agree with @danielbreves , this will mean people will need a big heads up before we make this change.

ocke: I was aware and also definitely agree with @danielbreves.

Unfortunately how the data was passed, isn't how it works in Promises 😢 .

What are you thinking when you say 'different strategy'.

ocke: I'm actually very impressed that we've had this bug since feb 2015 and nobody has really bumped into it since now.

danielbreves: This is coming from [jQuery's ajax](http://api.jquery.com/jquery.ajax/) [here](https://github.com/zendesk/zendesk_app_framework/blob/4cb0caaa644da761e4f32e2efb8f1eaa4c08833d/lib/request.js#L123-L126), right? I'd expect the first argument to be the payload body, no? I don't think there's an ideal solution, but maybe we can merge all arguments into one object?

ocke: @danielbreves That's a good point. I had a look again at the different of arguments recieved by a v1 and v2 app and saw a clear difference:

V1:
screen shot 2017-05-31 at 17 10 56

V2:
screen shot 2017-05-31 at 17 13 09

So what happens is that obviously `ajax.done` is returning the same arguments (3). But it's then send through `postMessage` and it strips that first argument away, cause that's a `#document` node. The second, a string, passes along just fine. The third, a object with some string and a bunch of functions, gets stripped down to only the strings and a number.

This also completely makes sense why it does work when using `data: text` because the first argument stays a string and doesn't get processed.

danielbreves: Nice, good investigation. In this case perhaps we could convert the node to text before `postMessage` then converting it back to a document node in ZAF SDK. That doesn't solve the problem that we're omitting a bunch of useful information, like request headers (as reported by others). Maybe we could make a minor bump for this :/

PR #47, opened 2017-07-26
[AF-785] accept object in invoke

ocke:
:v:

/cc @zendesk/vegemite

### Description
Accept object in invoke!

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-785

### Risks

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -487,7 +487,7 @@ Client.prototype = {
   get: function(path) {
     var paths = Array.isArray(path) ? path : [path];
 
-    if (arguments.length > 1 || paths.some(Utils.notIsString)) {
+    if (arguments.length > 1 || paths.some(function(path) { return !Utils.isString(path); })) {
       throw new Error('The get method accepts a string or array of strings.');
     }
 
@@ -523,7 +523,7 @@ Client.prototype = {
       Object.keys(keyOrObject).forEach(function(key) {
         var methodArgs = keyOrObject[key];
 
-        if (!Array.isArray(methodArgs) || methodArgs.some(Utils.notIsString)) {
+        if (!Array.isArray(methodArgs) || methodArgs.some(function(arg) { return !Utils.isString(arg); })) {
           throw new Error('Invoke supports string arguments or an object with array of strings.');
         }
       });
--- a/lib/client.js
+++ b/lib/client.js
@@ -4,7 +4,6 @@ var PROMISE_TIMEOUT = 5000, // 5 seconds
     version         = require('version'),
     Utils           = require('utils'),
     Promise         = window.Promise || require('../vendor/native-promise-only'),
-    when            = require('utils').when,
     pendingPromises = {},
     ids             = {};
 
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -514,13 +514,25 @@ Client.prototype = {
...
-    var obj = key;
+    var obj = {};
 
     if (typeof key === 'string') {
-      obj = {};
       obj[key] = Array.prototype.slice.call(arguments, 1);
+
+    } else if (typeof key === 'object') {
+      // Validate object
+      Object.keys(key).forEach(function(s) {
```
svizzari: `s` isn't an informative argument name 😾

spec/client_spec.js
```diff
@@ -794,6 +794,33 @@ describe('Client', function() {
             expect(lastCall.appGuid).to.equal('ABC123');
             expect(lastCall.instanceGuid).to.equal('def-321');
           });
+
+          it('makes a call with an object', function() {
+            promise = childClient.invoke({'popover': ['hide']});
```
svizzari: Do we need a test that has an unquoted key name?

ocke: that's the same thing in js

lib/client.js
```diff
@@ -514,13 +514,25 @@ Client.prototype = {
   },
 
   invoke: function(key) {
```
danielbreves: rename to `keyOrObject`?

lib/client.js
danielbreves: What's `s`?

lib/client.js
```diff
@@ -514,13 +514,25 @@ Client.prototype = {
...
+    var obj = {};
 
     if (typeof key === 'string') {
-      obj = {};
       obj[key] = Array.prototype.slice.call(arguments, 1);
+
+    } else if (typeof key === 'object') {
+      // Validate object
+      Object.keys(key).forEach(function(s) {
+        var values = key[s];
```
danielbreves: What's `values`? `methodArgs`?

lib/client.js
```diff
@@ -514,13 +514,25 @@ Client.prototype = {
...
+
+        if (!Array.isArray(values) || values.some(function(s) {return typeof s !== 'string'; })) {
```
danielbreves: Can we extract the second condition to a separate function?

lib/client.js
```diff
@@ -514,13 +514,25 @@ Client.prototype = {
...
+          throw new Error('Invoke supports string arguments or an object with array of strings.');
+        }
+      });
+
+      obj = key;
+
     } else {
-      throw new Error('Invoke only supports string arguments.');
+      throw new Error('Invoke supports string arguments or an object with array of strings.');
```
danielbreves: Can we extract this error message to a `var`?

lib/client.js
ocke: yeah i guess so

spec/client_spec.js
svizzari: > that's the same thing in js

yea, using quoted keys seems unconventional, why not just use unquoted

danielbreves:
@svizzari 💇‍♂️

PR #51, opened 2018-02-14
Export Promise

ocke:
/cc @zendesk/vegemite 

Currently we have a very nice polyfill in the SDK. But all apps that include this SDK, still don't have this polyfill available, which means all apps will need to include their own polyfill. This seems very strange to me, and it would be nice if we can export the polyfill to window, so apps automatically don't need to worry about Promise polyfill.

Export Promise to window so v2 apps don't need to include their own Promise polypill.

I have tested this, and it works great!

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -7,6 +7,9 @@ var PROMISE_TIMEOUT = 5000, // 5 seconds
     pendingPromises = {},
     ids             = {};
 
+// export Promise!
+window.Promise = Promise;
+
 // @internal
 // #timeoutReject(rejectionFn, name, params)
 //


lib/client.js
```diff
@@ -7,6 +7,9 @@ var PROMISE_TIMEOUT = 5000, // 5 seconds
     pendingPromises = {},
     ids             = {};
 
+// export Promise!
+window.Promise = Promise;
```
danielbreves: Could we namespace it under ZAFClient perhaps, to avoid making it seem like the browser supports it?

ocke: Isn't that the point of a polyfill? Making it so that it's like the browser supports it?

svizzari: It should be that the polyfill is present to provide a consistent API in v2 apps, but that the polyfill should no-op if the browser already supports Promise.  The polyfill we use is spec compliant, so it should be transparent

danielbreves: It just seems a bit intrusive to **implicitly** polyfill a native API, but I admit I can't think of any actual problems this could cause, so happy to 👍

danielbreves: It's possible developers are already overwriting this, so we shouldn't rely on it within our code.

svizzari: Should it be `window.Promise = window.Promise || Promise;`

ocke: @svizzari yes, but in the code we already to that:

```
Promise = window.Promise || require('../vendor/native-promise-only'),
(...)
window.Promise = Promise;
```

PR #52, opened 2018-02-14
[APPS-3789] Setup event listeners for app tracking

seangoedecke:
Relies on the ZAF work here https://github.com/zendesk/zendesk_app_framework/pull/769 to send tracking blips whenever someone clicks in an app.

:stop_sign: This should not be deployed until the corresponding ZAF PR above is deployed (apps won't break, but there'll be a lot of console errors as the `invoke` call fails to find a handler for `track`. :stop_sign: 

https://zendesk.atlassian.net/browse/APPS-3789

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -215,6 +215,11 @@ function processResponse(path, result) {
   return result;
 }
 
+function setupTracking(client) {
+  window.addEventListener('click', function() {
+    client.invoke('track', { type: 'click' }) // Arturo filtering is being done on the ZAF side
+}
+
 var Client = function(options) {
   this._parent = options.parent;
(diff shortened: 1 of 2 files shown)


seangoedecke:
@zendesk/vegemite @zendesk/wombat

lib/client.js
```diff
@@ -215,6 +215,12 @@ function processResponse(path, result) {
   return result;
 }
 
+function setupTracking(client) {
+  window.addEventListener('click', function() {
+    client.invoke('track', { type: 'click' }) // Arturo filtering is being done on the ZAF side
```
danielbreves: semicolon 👴

seangoedecke: I'm used to a linter autofixing that for me 😆

PR #53, opened 2018-03-01
optional queryString for init

ocke:
In jasmine we are trying to do fishy stuff to stub out the window.document.location.search, but instead we should be able to pass it to the init function so we can test this much easier!

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -24,8 +24,8 @@ var Client    = require('client'),
 ///   console.log('Hi ' + currentUser.name);
 /// });
 /// ```
-ZAFClient.init = function(callback) {
-  var queryParams = Utils.queryParameters(),
+ZAFClient.init = function(callback, queryString) {
+  var queryParams = Utils.queryParameters(queryString),
       hashParams = Utils.queryParameters(( document.location.hash || '' ).slice(1)),
       origin = queryParams.origin || hashParams.origin,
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -24,9 +24,10 @@ var Client    = require('client'),
 ///   console.log('Hi ' + currentUser.name);
 /// });
 /// ```
-ZAFClient.init = function(callback) {
-  var queryParams = Utils.queryParameters(),
-      hashParams = Utils.queryParameters(( document.location.hash || '' ).slice(1)),
+ZAFClient.init = function(callback, location) {
+  location = location || {};
```
danielbreves: `location` here is shadowing `window.location`. Can we call it something else?

ocke: so is `name` and `event`, but that doesn't seem to be a problem?

https://github.com/zendesk/zendesk_app_framework_sdk/blob/master/lib/client.js#L21
https://github.com/zendesk/zendesk_app_framework_sdk/blob/master/lib/client.js#L121

🤷‍♂️

ocke: what would you suggest we change it to?

danielbreves: Didn't know `name` was thing tbh. It's not critical, but would prefer to avoid it, especially when the real property is relevant in this context. Could change it to `loc` or `locationConfig`?

PR #55, opened 2018-03-05
[APPS-3789] Refactor tracking code to dodge bug in grunt-testem-mincer

seangoedecke:
`grunt-testem-mincer` is super out of date and relies on an old version of `testem`. This relies on an old version of `ws-pure`, which is forked from `ws` at the point where `ws` had a nasty bug (it would sometimes mysteriously throw out of range errors (https://github.com/websockets/ws/issues/778).

To avoid this bug I've just extracted all my code and stubbed it out in the test.

https://zendesk.atlassian.net/browse/APPS-3789

Risks:
none: just a test.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -5,7 +5,8 @@ var PROMISE_TIMEOUT = 5000, // 5 seconds
     Utils           = require('utils'),
     Promise         = window.Promise || require('../vendor/native-promise-only'),
     pendingPromises = {},
-    ids             = {};
+    ids             = {},
+    Tracking        = require('tracking');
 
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -252,7 +247,7 @@ var Client = function(options) {
     return this; // shortcut handshake
   }
 
-  setupTracking(this)
+  Tracking.setupTracking(this);
```
waferbaby: Would it make more sense to call this `Tracking.setup`?

seangoedecke: Sure! Done.

PR #56, opened 2018-03-19
allow multiple args in events

ocke:
/cc @zendesk/vegemite 

allow multiple args in events

https://github.com/zendesk/zendesk_app_framework/pull/773

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -124,7 +124,11 @@ function isValidEvent(client, event) {
 function triggerEvent(client, eventName, data) {
   if (!client._messageHandlers[eventName]) { return false; }
   client._messageHandlers[eventName].forEach(function(handler) {
-    handler(data);
+    if (Array.isArray(data)) {
+      handler.apply(undefined, data);
+    } else {
+      handler(data);
+    }
   });
 }
 


lib/client.js
```diff
@@ -124,7 +124,11 @@ function isValidEvent(client, event) {
 function triggerEvent(client, eventName, data) {
   if (!client._messageHandlers[eventName]) { return false; }
   client._messageHandlers[eventName].forEach(function(handler) {
-    handler(data);
+    if (Array.isArray(data)) {
+      handler.apply(undefined, data);
```
Mehul-Porwal: If we are using ES6, then you can simply make this `handler(...data)`

Also, is there any difference between passing `undefined` or `null` as the first argument to `apply`?

ocke: We are not useing es6 in the sdk.

Yes there is a difference., in the first `this` will be `undefined`, in the latter it will be `null`

lib/client.js
```diff
@@ -124,7 +124,11 @@ function isValidEvent(client, event) {
 function triggerEvent(client, eventName, data) {
   if (!client._messageHandlers[eventName]) { return false; }
   client._messageHandlers[eventName].forEach(function(handler) {
-    handler(data);
+    if (Array.isArray(data)) {
```
danielbreves: What is data was being sent as an array already? Could we send this from the framework using a special key to distinguish it further, like `responseArgs`?

e.g.
```
if (data.eventArgs) {
  handler.apply(undefined, data.eventArgs);
}
```

ocke: This can only happen to `zaf.` events. And I looked into framework code, this only happens for non-ajax requests, so if an ajax request returns an array, it won't hit this code path.

Do you know of any events that have an array as the first argument?

ocke: Actually, we already do this strange thing, where we wrap ajax data into `{ responseArgs: data}`, and then in the sdk, we unwrap it. Why is this?

danielbreves: Yes, that's what I meant. It's actually legacy from v1. But I don't think it's weird. It makes the intent more explicit.

danielbreves: The intent of passing arguments to the app's event handler, rather than any other kind of data that may be internal to the framework.

danielbreves: I don't see how the intent is wrong. The behaviour you described is the same I described: "passing arguments to the app's event handler", no? It's just not explicit. I can't tell just by reading this code what data is used for. 

As for over-engineering, I would agree with you if this was all one project in one repo. It would be easy to change the interface and consumers at the same time. This not being the case, I prefer to nest it so that if we need to pass additional internal properties it's easy to do without introducing temporary backwards compatibility code.

PR #57, opened 2018-03-20
[APPS-3792] Mouse hover length tracking

lcampbell-zendesk:
### Description

We will send tracking blips when a user hovers over an app for more than 200ms. This minimum time is to limit the number of blips generated in instances where a user might be wiggling the mouse around.

### References

* JIRA: [As an agent, when I move my mouse out of a sidebar app, a blip should be sent to Hive
](https://zendesk.atlassian.net/browse/APPS-3792)

### Risks

* [low] May generate as many as 5 per second per user

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,6 @@
 language: node_js
 node_js:
   - '6.3.0'
-before_install:
-  - export PATH=$PWD/travis_phantomjs/phantomjs-$PHANTOMJS_VERSION-linux-x86_64/bin:$PATH
-  - if [ $(phantomjs --version) != '$PHANTOMJS_VERSION' ]; then rm -rf $PWD/travis_phantomjs; mkdir -p $PWD/travis_phantomjs; fi
-  - if [ $(phantomjs --version) != '$PHANTOMJS_VERSION' ]; then wget https://github.com/Medium/phantomjs/releases/download/v$PHANTOMJS_VERSION/phantomjs-$PHANTOMJS_VERSION-linux-x86_64.tar.bz2 -O $PWD/travis_phantomjs/phantomjs-$PHANTOMJS_VERSION-linux-x86_64.tar.bz2; fi
-  - if [ $(phantomjs --version) != '$PHANTOMJS_VERSION' ]; then tar -xvf $PWD/travis_phantomjs/phantomjs-$PHANTOMJS_VERSION-linux-x86_64.tar.bz2 -C $PWD/travis_phantomjs; fi
-  - phantomjs --version
-before_script:
-  - npm install -g grunt-cli
 sudo: false
 cache:
   directories:
@@ -16,5 +8,3 @@ cache:
     - travis_phantomjs
 branches:
   only: [master]
-env:
-  - PHANTOMJS_VERSION=2.1.1
--- a/lib/client.js
+++ b/lib/client.js
@@ -6,7 +6,9 @@ var PROMISE_TIMEOUT = 5000, // 5 seconds
     Promise         = window.Promise || require('../vendor/native-promise-only'),
     pendingPromises = {},
     ids             = {},
-    Tracking        = require('tracking');
+    Tracking        = require('tracking'),
+    Tracker         = Tracking.Tracker;
+
 
 // export Promise!
 window.Promise = Promise;
@@ -220,6 +222,7 @@ function processResponse(path, result) {
 }
 
 var Client = function(options) {
+  var tracker = new Tracker(this);
   this._parent = options.parent;
   this._origin = options.origin || this._parent && this._parent._origin;
   this._source = options.source || this._parent && this._parent._source || window.parent;
@@ -247,7 +250,7 @@ var Client = function(options) {
     return this; // shortcut handshake
   }
 
-  Tracking.setup(this);
+  tracker.setup();
   window.addEventListener('message', messageHandler.bind(null, this));
   this.postMessage('iframe.handshake', { version: version });
 };
--- a/lib/client.js
+++ b/lib/client.js
@@ -6,9 +6,7 @@ var PROMISE_TIMEOUT = 5000, // 5 seconds
     Promise         = window.Promise || require('../vendor/native-promise-only'),
     pendingPromises = {},
     ids             = {},
-    Tracking        = require('tracking'),
-    Tracker         = Tracking.Tracker;
-
+    Tracker         = require('tracking').Tracker;
 
 // export Promise!
 window.Promise = Promise;
@@ -222,7 +220,6 @@ function processResponse(path, result) {
 }
 
 var Client = function(options) {
-  var tracker = new Tracker(this);
   this._parent = options.parent;
   this._origin = options.origin || this._parent && this._parent._origin;
   this._source = options.source || this._parent && this._parent._source || window.parent;
@@ -250,6 +247,7 @@ var Client = function(options) {
     return this; // shortcut handshake
   }
 
+  var tracker = new Tracker(this);
   tracker.setup();
   window.addEventListener('message', messageHandler.bind(null, this));
   this.postMessage('iframe.handshake', { version: version });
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -1,20 +1,10 @@
 language: node_js
 node_js:
   - '6.3.0'
-before_install:
-  - export PATH=$PWD/travis_phantomjs/phantomjs-$PHANTOMJS_VERSION-linux-x86_64/bin:$PATH
-  - if [ $(phantomjs --version) != '$PHANTOMJS_VERSION' ]; then rm -rf $PWD/travis_phantomjs; mkdir -p $PWD/travis_phantomjs; fi
-  - if [ $(phantomjs --version) != '$PHANTOMJS_VERSION' ]; then wget https://github.com/Medium/phantomjs/releases/download/v$PHANTOMJS_VERSION/phantomjs-$PHANTOMJS_VERSION-linux-x86_64.tar.bz2 -O $PWD/travis_phantomjs/phantomjs-$PHANTOMJS_VERSION-linux-x86_64.tar.bz2; fi
-  - if [ $(phantomjs --version) != '$PHANTOMJS_VERSION' ]; then tar -xvf $PWD/travis_phantomjs/phantomjs-$PHANTOMJS_VERSION-linux-x86_64.tar.bz2 -C $PWD/travis_phantomjs; fi
-  - phantomjs --version
```
lcampbell-zendesk: We don't need any of this stuff?

lib/tracking.js
```diff
@@ -1,7 +1,29 @@
...
+    overStart = Date.now();
+  });
+
+  window.addEventListener('mouseout', function() {
+    var overFor = Date.now() - overStart;
+    if(overFor >= MIN_HOVER_TIME) {
+      client.invoke('track', { type: 'hover',  hoverLength: overFor });
+    }
+  });
+}
```
lcampbell-zendesk: Good code. I like this code.

spec/tracking_spec.js
```diff
@@ -0,0 +1,92 @@
+describe('App Tracking', function() {
+  var tracking = require('tracking'),
+      sandbox  = sinon.sandbox.create(),
```
lcampbell-zendesk: Where does `sinon` come from?

spec/tracking_spec.js
```diff
@@ -0,0 +1,92 @@
...
+    sandbox.restore();
+  });
+
+  function findEvent(event) {
+    for(var i = 0; i < eventListener.args.length; i++) {
+      if(eventListener.args[i][0] === event) {
+        return eventListener.args[i];
+      }
+    }
+  }
```
lcampbell-zendesk: This is fragile. It's possible for multiple listeners of a single type to be defined. This method might grab the wrong one and cause tests to fail despite the code working fine. I might experiment with looking for events based on function name (you'd need to give them function names with `function hover(a,b,c) {...}`). Or not test the event registration at all. Instead pulling the callbacks out into their own named functions and testing their logic as unit tests with none of the need for the `beforeEach` and `afterEach` or `findEvent` stuff that you have here.

spec/tracking_spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+  describe('for click events', function() {
+    it('attaches the click event listener', function() {
+      var clickEvent = findEvent('click');
+
+      expect(clickEvent).not.to.be.undefined;
```
lcampbell-zendesk: Does this expectation short circuit the whole it block? If not then if this fails it will cause exceptions on the next two lines.

spec/tracking_spec.js
```diff
@@ -0,0 +1,92 @@
...
+      expect(mouseoverEvent[0]).to.equal('mouseover');
+      expect(mouseoverEvent[1]).to.be.a.function;
+    });
+
+    it('attaches the mousout event listener', function() {
+      var mouseoutEvent = findEvent('mouseout');
+      
+      expect(mouseoutEvent).not.to.be.undefined;
+      expect(mouseoutEvent[0]).to.equal('mouseout');
+      expect(mouseoutEvent[1]).to.be.a.function;
```
lcampbell-zendesk: Test code looks different to normal code. It's easy to fall out of the good habits that we follow for normal code. Here is the third repeat of these three lines of code:

``` javascript
expect(mouseoutEvent).not.to.be.undefined;
expect(mouseoutEvent[0]).to.equal('mouseout');
expect(mouseoutEvent[1]).to.be.a.function;
```

I suggest pulling it out into a helpfully named function and parameterising it based on the event type. Honestly you could probably pull in

``` javascript
var mouseoutEvent = findEvent('mouseout');
```

as well. Maybe `function expectEventToBeRegistered(eventType) { ... }`.

spec/tracking_spec.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    context('when the mouseout event is triggered', function() {
+      var mouseoverCallback, mouseoutCallback, clock;
+
+      function causeMouseEvents(time) {
+        mouseoverCallback();
+        clock.tick(time);
+        mouseoutCallback();
+      }
```
lcampbell-zendesk: Good that you pulled this out into a method.

.travis.yml
StaphSynth: Nope. I added phantomjs to the dev dependencies in `package.json`, obviating this logic (it was actually this logic that broke the build because it made hard-coded assumptions about the PJS version).

spec/tracking_spec.js
StaphSynth: [Sinon](http://sinonjs.org/) is the stubs/mocks/spies library this uses.

lib/tracking.js
lcampbell-zendesk: If you changed lines 10-16 like so:

``` javascript
function hoverHandler() {
  var overFor = Date.now() - overStart;
  if(overFor >= MIN_HOVER_TIME) {
    client.invoke('track', { type: 'hover',  hoverLength: overFor });
  }
}

window.addEventListener('mouseout', hoverHandler);
```

Then you could refer to `tracking.hoverHandler` directly in the tests without needing to use `findEvent` to look for it on the `eventListener`.

spec/tracking_spec.js
lcampbell-zendesk: And it's available without needing `require`? Odd.

lib/client.js
```diff
@@ -247,7 +250,7 @@ var Client = function(options) {
     return this; // shortcut handshake
   }
 
-  Tracking.setup(this);
+  tracker.setup();
```
seangoedecke: nit: I wonder if it's better to bring the initialization on line 225 down here, so that they're in the same spot

lib/tracking.js
```diff
@@ -1,7 +1,32 @@
...
-module.exports = { setup: setup };
+Tracker.prototype = {
+  handleMouseOver: function() {
+    this.startTime = Date.now();
+  },
+
+  handleMouseOut: function() {
+    var overFor = Date.now() - this.startTime;
+    if(overFor >= this.MIN_HOVER_TIME) {
+      this.client.invoke('track', { type: 'hover', hoverLength: overFor });
```
seangoedecke: Should this be `value: overFor`?

spec/client_spec.js
```diff
@@ -1,17 +1,19 @@
 describe('Client', function() {
   var Client   = require('client'),
       Tracking = require('tracking'),
+      Tracker  = Tracking.Tracker,
```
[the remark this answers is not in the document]
StaphSynth: Good idea. Done.

lib/tracking.js
```diff
@@ -1,7 +1,32 @@
-function setup(client) {
-  window.addEventListener('click', function() {
-    client.invoke('track', { type: 'click' });
-  });
+function Tracker(client) {
```
ocke: nit: shouldn't this be `tracker.js`?

StaphSynth: 🤷‍♂️ I didn't name the file.

PR #58, opened 2018-03-20
Add a whitelist of origins

Mehul-Porwal:
Any website can include zendesk_app_framework, embed one of our iframed app, and be able to sniff the events and data. This PR is adding a whitelist of domains to prevent this.

Reported Issue: https://support.zendesk.com/agent/tickets/3371584

@zendesk/vegemite 

Limitations: Since this is a front-end only fix, anyone can easily bypass this, we should think about adding nginx rules on the app assets to not be allowed to be iframed on any other domains.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -219,6 +219,26 @@ function processResponse(path, result) {
   return result;
 }
 
+function isOriginValid(origin) {
+  var WHITELISTED_ORIGINS = [
+    /.*zendesk\.com.*/gi,
+    /.*zd-staging\.com.*/gi,
+    /.*zd-dev\.com.*/gi,
+    /.*zd-master\.com.*/gi,
+    /.*zendesk-staging\.com.*/gi,
+    /.*dashboard\.zopim\.com.*/gi,
+    /.*dashboard\.zopim\.org.*/gi
+  ];
+
+  for (var i = 0; i < WHITELISTED_ORIGINS.length; i++) {
+    if (WHITELISTED_ORIGINS[i].test(origin)) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
 var Client = function(options) {
   this._parent = options.parent;
   this._origin = options.origin || this._parent && this._parent._origin;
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -232,6 +232,28 @@ var Client = function(options) {
...
+    /.*zendesk\.com.*/gi,
+    /.*zd-staging\.com.*/gi,
+    /.*zd-dev\.com.*/gi,
+    /.*zd-master\.com.*/gi,
+    /.*zendesk-staging\.com.*/gi,
+    /.*dashboard\.zopim\.com.*/gi,
+    /.*dashboard\.zopim\.org.*/gi
+  ];
+
+  var i;
```
svizzari: Couldn't just declare the iterator var in the `for` statement? Seems unnecessary to pull it out.  If you _were_ going to do that, I'd expect you might save the length of the Array in the constant into another var, like `len = WHITELISTED_ORIGINS.length`.

lib/client.js
```diff
@@ -232,6 +232,28 @@ var Client = function(options) {
   this._context = options.context || null;
   this.ready = false;
 
+  var WHITELISTED_ORIGINS = [
```
svizzari: I'm guessing it was a conscious choice to make the Regexp literals simple so as to be unambiguous?  (Could have put all of them into a single Regexp with optional groups, then just avoided the loop below)

lib/client.js
```diff
@@ -232,6 +232,28 @@ var Client = function(options) {
...
+    /.*dashboard\.zopim\.org.*/gi
+  ];
+
+  var i;
+
+  for (i = 0; i < WHITELISTED_ORIGINS.length; i++) {
+    if (WHITELISTED_ORIGINS[i].test(this._origin)) {
+      break;
+    }
+  }
```
danielbreves: Could we extract this into a function that returns `true` if the origin passed matches any regex, then if the function returns `true` throw the exception below?

danielbreves: Could we add a test please?

Mehul-Porwal: I'll move it to a separate function. I think I also need to add `*.apps.zdusercontent.com` to the list.

Mehul-Porwal:
@svizzari @danielbreves Please help review again. 🙏

spec/client_spec.js
```diff
@@ -36,6 +36,85 @@ describe('Client', function() {
     window.addEventListener.callArgWith(1, evt);
   }
 
+  describe('isDomainValid', function() {
+    it('Should instantiate client when domain is valid 1', function() {
```
danielbreves: Could we refer to the domain in the description, rather than a count, just so it's easier understand when they fail?

spec/client_spec.js
```diff
@@ -36,6 +36,78 @@ describe('Client', function() {
     window.addEventListener.callArgWith(1, evt);
   }
 
+  describe('isDomainValid', function() {
```
svizzari: `isOriginValid` to match the helper method (intent)? Or, possibly just change to:

```js
describe("with a valid origin", () => {})
describe("with an invalid origin", () => {})
```

?

Mehul-Porwal:
@svizzari @danielbreves Do you know why the tests are failing?

danielbreves:
Seems like it's failing on master too: https://travis-ci.org/zendesk/zendesk_app_framework_sdk/builds/359212069

Mehul-Porwal:
@token-cjg Can I request you to QA this PR, it is already deployed to Staging and Master. The first step is to verify that all apps should continue to function normally in all stages including Master, Staging and Production. Second step should be to test the use case mentioned in https://support.zendesk.com/agent/tickets/3371584.

token-cjg:
@zendesk/wombat, might be similar to https://github.com/zendesk/zendesk_app_market/pull/2737 and https://github.com/zendesk/zendesk_app_market/pull/2743 in impact re: host-mapped accounts.

@Mehul-Porwal, I think you've reverted this now and merged the revert into master, yes?

PR #60, opened 2018-05-07
Revert "Revert "Add a whitelist of origins""

danielbreves:
Reverts zendesk/zendesk_app_framework_sdk#59


token-cjg:
:+1:

Tested against a pod8 hostmapped account with an app pointing to the corresponding release of the sdk in staging.

PR #63, opened 2018-05-08
Fix regex (take 3)

### LINKED ISSUE #62: isOriginValid function in v2.0.14 does not seem to accept hyphenated urls
My url is something like `https://dev-mycomp.zendesk.com` This url was working fine previously for me but suddenly after the latest update for v2.0.14, my app have stopped working and it returns **Uncaught Error: Invalid domain** .  The regex'es used in [isOriginValid()](https://github.com/zendesk/zendesk_app_framework_sdk/blob/0f0e96cb3113f835ecb5d941f5a98d442b5d76c4/lib/client.js#L222) seems to be missing some extra patterns. 
Could you pls look into it.

danielbreves:
Fixes #62

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -223,12 +223,12 @@ function isOriginValid(origin) {
   var WHITELISTED_ORIGINS = [
     /^http(s)?:\/\/127.0.0.1(:\d+)?$/,
     /^http(s)?:\/\/localhost(:\d+)?$/,
-    /^https:\/\/\w+\.zendesk\.com$/,
-    /^https:\/\/\w+\.zd-staging\.com$/,
-    /^https:\/\/\w+\.zd-dev\.com$/,
-    /^https:\/\/\w+\.zd-master\.com$/,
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -221,14 +221,14 @@ function processResponse(path, result) {
 
 function isOriginValid(origin) {
   var WHITELISTED_ORIGINS = [
-    /^http(s)?:\/\/127.0.0.1(:\d+)?$/,
-    /^http(s)?:\/\/localhost(:\d+)?$/,
-    /^https:\/\/\w+\.zendesk\.com$/,
```
ocke: just wondering what's wrong with `\w` cause it means letter, number, underscore. sounds like a url to me. `.` will also match spaces etc.

danielbreves: It doesn't work with dashes, and who knows what else...

ocke: who has dashes in their url!?!?!?! :trollface:

PR #65, opened 2018-06-07
Add apps.zdusercontent.com to whitelisted domains

Mehul-Porwal:
We missed out one domain in our whitelist, adding it in here.

@zendesk/vegemite 

## ToDo

- [x] Add test for this domain

## Risk
Low. Just one more whitelist domain regex


danielbreves:
@Mehul-Porwal why would apps.zdusercontent.com iframe an app?

PR #66, opened 2018-06-14
Instead of throw, send to DD

Mehul-Porwal:
Send the invalid domain error to Datadog instead of throwing. We can monitor this for a few days before we release the versions which throws, as throwing an error here(if done wrong) can we very high risk and can break a lot of apps.

Requires ZAF PR: https://github.com/zendesk/zendesk_app_framework/pull/788

@zendesk/vegemite

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -254,8 +254,15 @@ var Client = function(options) {
   this._context = options.context || null;
   this.ready = false;
 
-  if (!isOriginValid(this._origin)) {
-    throw new Error('Invalid domain ' + this._origin);
+  if (this._origin && isOriginValid(this._origin)) {
+    var origin_hostname = new URL(this._origin).hostname;
+    this.postMessage('tracker', {
+      event_name: 'invalid_domain',
+      event_type: 'increment',
+      data: 1, 
+      tags: ["origin:" + origin_hostname]
+    });
+    // throw new Error('Invalid domain ' + this._origin);
   }
 
   this.on('app.registered', function(data) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -220,21 +220,23 @@ function processResponse(path, result) {
 }
 
 function isOriginValid(origin) {
-  var WHITELISTED_ORIGINS = [
-    /^https?:\/\/127.0.0.1(:\d+)?$/,
-    /^https?:\/\/localhost(:\d+)?$/,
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -254,8 +254,15 @@ var Client = function(options) {
...
-    throw new Error('Invalid domain ' + this._origin);
+  if (this._origin && isOriginValid(this._origin)) {
+    var origin_hostname = new URL(this._origin).hostname;
+    this.postMessage('tracker', {
+      event_name: 'invalid_domain',
+      event_type: 'increment',
+      data: 1, 
+      tags: ["origin:" + origin_hostname]
+    });
+    // throw new Error('Invalid domain ' + this._origin);
```
danielbreves: Can we leave this comment out?

lib/client.js
```diff
@@ -254,8 +254,15 @@ var Client = function(options) {
   this._context = options.context || null;
   this.ready = false;
 
-  if (!isOriginValid(this._origin)) {
-    throw new Error('Invalid domain ' + this._origin);
+  if (this._origin && isOriginValid(this._origin)) {
```
danielbreves: Can we move the guard to within the `isOriginValid` function?

lib/client.js
```diff
@@ -254,8 +254,15 @@ var Client = function(options) {
...
+    var origin_hostname = new URL(this._origin).hostname;
+    this.postMessage('tracker', {
```
danielbreves: Wondering if we should make this message a little less guessable, just to avoid accidental calls? e.g. `__track__`

lib/client.js
```diff
@@ -254,8 +254,15 @@ var Client = function(options) {
...
+      event_name: 'invalid_domain',
```
danielbreves: `invalid_sdk_origin`?

Mehul-Porwal:
@danielbreves Can you have a  👀 again?

hoangtrvu:
Did you create a new tag for this @Mehul-Porwal? See https://github.com/zendesk/zaf_sdk_deployment/blob/master/DEPLOY.md#deployment-process. Looks like master should have latest tag at v2.0.17 now, v2.0.16 is behind and doesn't include this PR.

Mehul-Porwal:
@hoangtrvu Done! Created the tag.

PR #71, opened 2018-08-27
2018 tools update

ocke:
/cc @zendesk/vegemite @zendesk/apps-migration 

Bring the SDK to modern levels with old tools removed and new tools in.

Go by commits for easy reading.

## Replacements
- eslint => standard
- grunt => webpack
- bower => npm/webpack
- testem => karma/assert/sinon
- lib/vendor/native-promise-only.js from bower => npm
- `public/` test app. Not really sure what this was used for. It looks like a initial way of testing the sdk, but we should rely on actual tests and by now we can simply run the sdk in any app by doing `npm run watch`. So it feels a bit useless.

## Stats
old file: 18,563 byte
new file: 20,123 byte

## Installing
old install: 18 seconds
new install: 22 seconds

## Building
old build: 1 second
new build: 3 seconds

## Image
screen shot 2018-09-21 at 11 56 14

`node_modules/` is native_promises.

So it's slower and bigger 😆. But I don't think that's a big deal.


test_app/manifest.json
```diff
@@ -1,19 +0,0 @@
-{
```
ocke: leave this in.

.gitignore
```diff
@@ -1,7 +1,7 @@
...
-/node_modules/
-/bower_components/
-/build
+node_modules/
+bower_components/
+dist/
 vendor/bundle
 npm-debug.log
-/tmp
+tmp
```
ocke: add `/`

.gitignore
```diff
@@ -1,7 +1,7 @@
 .bundle
-/node_modules/
-/bower_components/
-/build
+node_modules/
+bower_components/
+dist/
```
ocke: capfile cares about this?

ocke:
fix stats and post stats in PR

lib/utils.js
```diff
@@ -57,73 +52,62 @@ function isString(key) {
...
-    resolveAll();
-    return allResolvedPromise;
+    resolveAll()
+    return allResolvedPromise
   }
 
-  itemsCopy.forEach(function(item, index) {
-    var promise;
+  itemsCopy.forEach((item, index) => {
+    var promise
```
keitheous: I noticed a few `var`s - was wondering if it is worthwhile to convert these to `let`s, including the one in the for loop?

Maybe we can do a for(const of) loop?

Mehul-Porwal:
This seems like a lot of changes and high risk. Is there anyway we can break all these changes into smaller PRs? 1 PR for grunt => webpack, another one for moving from bower to npm, etc.

Also I'm not a big fan of standard, as it cannot be customized and the no semi-colons thing is not my personal favourite.

.travis.yml
```diff
@@ -1,10 +1,19 @@
 language: node_js
 node_js:
   - '6.3.0'
```
yu-tian113: webpack 4 requires node >= 6.3.15

lib/utils.js
```diff
@@ -1,53 +1,48 @@
...
 
-  keyValuePairs = queryString.split('&');
+  keyValuePairs = queryString.split('&')
 
-  for (var i = 0; i < keyValuePairs.length; i++) {
-    keyAndValue = keyValuePairs[i].split('=');
-    key   = decode(keyAndValue[0]);
-    value = decode(keyAndValue[1]) || '';
-    result[key] = value;
+  for (let i = 0; i < keyValuePairs.length; i++) {
```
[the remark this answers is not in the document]
Mehul-Porwal: linter will not allow you to do this `for (let i = 0; i < keyValuePairs.length; i++) {`.

It will complain that the variable is never reassigned, please use `const`.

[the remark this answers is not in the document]
ocke: `i++` ?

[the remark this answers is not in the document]
keitheous: Ops. I think Mehul and I got confused. We can't use const when iterating with this old school for loop

HOWEVER, the other for loops. I believe we were confused for in and for of.

PR #73, opened 2018-10-19
Grunt and testem => Webpack and karma

ocke:

This has no code changes (except for the require path!) and switches 

grunt -> webpack
testem -> karma

**review by commit!**


package.json
```diff
@@ -28,27 +28,31 @@
...
-    "grunt-eslint": "^19.0.0",
-    "grunt-gluejs2": "^0.3.0",
-    "grunt-newer": "^1.1.2",
-    "grunt-testem-mincer": "^0.5.18",
-    "load-grunt-tasks": "^3.5.0",
-    "phantomjs": "^2.1.7",
-    "sinon": "^1.17.3",
-    "sinon-chai": "^2.8.0",
-    "time-grunt": "^1.3.0"
+    "dirty-chai": "^2.0.1",
```
danielbreves: Nice!

package.json
```diff
@@ -28,27 +28,31 @@
     "framework",
     "sdk"
   ],
+  "dependencies": {
+    "native-promise-only": "^0.8.1",
+    "webpack": "^4.19.0",
+    "webpack-cli": "^3.1.0"
```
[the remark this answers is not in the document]
ocke: As discussed in 🍖 space, this is a little optimising apps team has used to be able to deploy faster on samson. When you only install the `dependencies` you can build the javascript. All `devDependencies` aren't strictly needed to build the js.

lib/index.js
```diff
@@ -43,4 +43,5 @@ ZAFClient.init = function(callback, loc) {
   return client;
 };
 
+window.ZAFClient = ZAFClient;
```
token-cjg: Would it be possible to add a unit test for this functionality?

danielbreves: Shouldn't we be using https://webpack.js.org/configuration/output/#output-library?

ocke: I would love to use it, but I can't get it to work.

ocke: I removed this in the next PR.

PR #74, opened 2018-10-20
eslint => standard

ocke:

This has no code changes, but a lot of linter code changes. Syntax was changed, never any code meaning.

standard

**review by commit!**


lib/index.js
```diff
@@ -1,46 +1,54 @@
...
+  loc = loc || window.location
+  var queryParams = Utils.queryParameters(loc.search)
+
+  var hashParams = Utils.queryParameters(loc.hash)
+
+  var origin = queryParams.origin || hashParams.origin
+
+  var appGuid = queryParams.app_guid || hashParams.app_guid
+
+  var client
```
danielbreves: Why so many line breaks?

lib/utils.js
```diff
@@ -1,53 +1,57 @@
...
-      value;
+  var result = {}
+
+  var keyValuePairs
+
+  var keyAndValue
+
+  var key
+
+  var value
```
danielbreves: Why so many line breaks?

lib/index.js
ocke: yeah i know, i have no idea why standard did that! This was after `standard --fix` 🤷‍♂️

ocke: I removed these linebreaks.

lib/utils.js
ocke: I removed the added linebreaks.

ocke:
rebased on master.

PR #75, opened 2018-10-20
ES6

ocke:

ES6 code!

**review by (only 3) commits**

### es6
es6


lib/index.js
```diff
@@ -22,33 +19,30 @@ var ZAFClient = {}
...
+    }
+
+    return client
   }
-
-  return client
 }
 
-module.exports = ZAFClient
+window.ZAFClient = ZAFClient
```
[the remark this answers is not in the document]
ocke: So others can use it?

webpack.config.js
```diff
@@ -11,13 +11,15 @@ module.exports = function (env = {}) {
       'zaf_sdk': [
         'native-promise-only',
         './lib/index.js'
-      ],
```
ocke: I moved this to the `if production` condition, because for tests this was taking forever.

spec/index_spec.js
```diff
@@ -1,50 +1,18 @@
...
-      describe('when a callback is passed', function () {
-        var callback = function () { return 'abcxyz' }
-        callback.bind = function () { return callback }
-
-        beforeEach(function () {
-          sandbox.spy(Client.prototype, 'on')
-          ZAFClient.init(callback)
-        })
-
-        it('binds the callback to app.registered', function () {
```
danielbreves: Where did these tests go?

ocke: they are back! 🎉

spec/utils_spec.js
```diff
@@ -1,12 +1,12 @@
 /* eslint-env mocha */
-/* global expect */
-describe('Utils', function () {
-  var Promise = window.Promise || require('native-promise-only')
```
hoangtrvu: Do we still need to import Promise here?

webpack.config.js
```diff
@@ -45,5 +47,24 @@ module.exports = function (env = {}) {
...
 
+  if (env.production) {
+    config.entry['zaf_sdk.min'] = [
+      'native-promise-only',
+      './lib/index.js'
+    ]
+
+    config.module.rules.push({
+      test: /\.js$/,
+      use: { loader: 'babel-loader', options: { plugins: [], presets: ['babel-preset-env'] } }
```
hoangtrvu: Should we use babel for development as well, for example, to test with various browser/os?

spec/utils_spec.js
ocke: I did a bit of digging, but that line is very old from when we were running in phantom, not in chrome. We only run in chrome now, so Promise will always be available.

lib/client.js
```diff
@@ -240,7 +241,7 @@ function isOriginValid (origin) {
       /^https:\/\/dashboard\.zopim\.org$/
     ]
 
-    for (var i = 0; i < WHITELISTED_ORIGINS.length; i++) {
+    for (let i = 0; i < WHITELISTED_ORIGINS.length; i++) {
```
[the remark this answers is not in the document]
ocke: as discussed.

PR #77, opened 2018-10-31
Send data to DD on timeout reject

Mehul-Porwal:
@zendesk/vegemite 

Send an event to DD when a ZAF request times out.

### Risk
LOW. Just adding a new metric to send to DD.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -17,7 +17,7 @@ window.Promise = Promise;
 // Reject a request if required, given the rejection function, name (get, set or invoke)
 // and arguments to that request. Falls back to 5000 second timeout with few exceptions.
 //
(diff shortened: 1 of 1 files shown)


Mehul-Porwal:
@fvzzy Sorry, it was not the latest code change. Just pushed one change on top of it.

`this.postMessage` -> `this.zafClient.postMessage`

lib/client.js
```diff
@@ -42,6 +42,12 @@ function timeoutReject(reject, name, paramsArray) {
 
 function defaultTimer(callback) {
   return setTimeout(function() {
+    this.zafClient.postMessage('__track__', {
```
hoangtrvu: I'm not sure how `zafClient` is set to `window`?

PR #81, opened 2018-11-20
Increase promise timeout to 10 seconds

danielbreves:
@zendesk/vegemite

Increase promise timeout to 10 seconds to allow more time for requests to resolve. It seems especially during initialisation 5 seconds may not be enough for requests to complete when many apps are loading at the same time.

### References
ZD#4058685

### Risks
- [low]  In case there's a legitimate reason for `postMessage` to never resolve it will take longer for the app to be able to handle the error.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1,5 +1,5 @@
-var PROMISE_TIMEOUT = 5000
-// 5 seconds
+var PROMISE_TIMEOUT = 10000
+// 10 seconds
 var PROMISE_DONT_TIMEOUT = ['instances.create']
 var ZAF_EVENT = /^zaf\./
 var version = require('version')
--- a/lib/client.js
+++ b/lib/client.js
@@ -1,5 +1,5 @@
 var PROMISE_TIMEOUT = 10000
-// 10 seconds
+// 10 seconds, see ZD#4058685
 var PROMISE_DONT_TIMEOUT = ['instances.create']
 var ZAF_EVENT = /^zaf\./
 var version = require('version')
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -1,5 +1,5 @@
-var PROMISE_TIMEOUT = 5000
-// 5 seconds
+var PROMISE_TIMEOUT = 10000
+// 10 seconds
```
hoangtrvu: Let's reference the ticket here so we have some context if next time we consider changing this again.

bryan-flynn-zd:
Hi @danielbreves @Mehul-Porwal  -- when will this be rolled out? Coherence via ticket [https://support.zendesk.com/agent/tickets/4058685](https://support.zendesk.com/agent/tickets/4058685) really wants this as soon as they can get it. 

I'm assuming, too, that existing zaf_sdk assets will be updated in place, correct? (IOW, Coherence won't have to reference a different SDK version or location)

hoangtrvu:
Hi @bryan-flynn-zd, yes that's correct, they won't have to update their reference to the SDK. We're trying to see if we could deploy this today.

bryan-flynn-zd:
Hi @hoangtrvu -- I see the deploy but when I look at the actual asset files, I still see the 5000ms timeout value in both the older and newer locations:
[https://static.zdassets.com/zendesk_app_framework_sdk/2.0/zaf_sdk.js](https://static.zdassets.com/zendesk_app_framework_sdk/2.0/zaf_sdk.js)
[https://static.zdassets.com/zendesk_app_framework_sdk/2.0/zaf_sdk.min.js](https://static.zdassets.com/zendesk_app_framework_sdk/2.0/zaf_sdk.min.js)

Am I missing something?

Mehul-Porwal:
@bryan-flynn-zd I think it is just a browser cache. Opening it in incognito gives this:

screen shot 2018-12-04 at 10 32 31 am

PR #85, opened 2019-01-22
[AF-1476] Add API tags to request timeout tracking

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
Adds tags to the `sdk_request_timeout` event tracking to capture which APIs timed out. 

Related to https://github.com/zendesk/zendesk_app_framework/pull/823.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1476

### Risks
* Low - Datadog tracking for `sdk_request_timeout` fails because of errors in tag strings.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -16,12 +16,13 @@ const ids = {}
 window.Promise = Promise
 
 // @internal
-// #timeoutReject(rejectionFn, name, params)
+// #timeoutReject(rejectionFn, name, client, params)
 //
 // Reject a request if required, given the rejection function, name (get, set or invoke)
 // and arguments to that request. Falls back to 10000 second timeout with few exceptions.
 //
 function timeoutReject (reject, name, client, paramsArray) {
+  const actions = paramsArray.map(action => `${name}-${action}`)
   switch (name) {
     case 'invoke': {
       const matches = paramsArray.filter((action) => {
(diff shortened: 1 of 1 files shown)


lib/client.js
```diff
@@ -35,21 +36,23 @@ function timeoutReject (reject, name, client, paramsArray) {
...
     }
     default: {
-      return defaultTimer(client, reject)
+      return defaultTimer(client, reject, actions)
     }
   }
 }
 
-function defaultTimer (client, callback) {
+function defaultTimer (client, callback, actions) {
```
Mehul-Porwal: Can we leave the callback as the last argument? It helps make the code more readable. If callback is the last argument then we can also pass in a inline anonymous function as callback and still have readable code.

fvzzy: Yep good suggestion 👍

lib/client.js
```diff
@@ -16,12 +16,13 @@ const ids = {}
...
 
 // @internal
-// #timeoutReject(rejectionFn, name, params)
+// #timeoutReject(rejectionFn, name, client, params)
 //
 // Reject a request if required, given the rejection function, name (get, set or invoke)
 // and arguments to that request. Falls back to 10000 second timeout with few exceptions.
 //
 function timeoutReject (reject, name, client, paramsArray) {
+  const actions = paramsArray.map(action => name + '-' + action)
```
ocke: Suggestion: `${name}-${action}`

lib/client.js
```diff
@@ -35,21 +36,23 @@ function timeoutReject (reject, name, client, paramsArray) {
...
     default: {
-      return defaultTimer(client, reject)
+      return defaultTimer(client, actions, reject)
     }
   }
 }
 
-function defaultTimer (client, callback) {
+function defaultTimer (client, actions, callback) {
+  const actionsTags = actions.map(action => 'action:' + action)
```
ocke: Suggestion: `action: ${action}`

PR #88, opened 2019-03-08
[AF-1528] Add instrumentation to measure time for SDK request timeouts

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
Extends the SDK request timeout to 5 minutes, and adds instrumentation to track the time taken to resolve every request longer than 10 seconds.

Measured times are grouped into 10 second ranges to ensure tags sent to Datadog are bounded.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1528

### Risks
* Low - Datadog tracking for `sdk_request_timeout` fails because of errors in tag strings.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -7,7 +7,7 @@ import NativePromise from 'native-promise-only'
 const Promise = window.Promise || NativePromise
 const PROMISE_TIMEOUT = 10000
 // 10 seconds, see ZD#4058685
-const PROMISE_DONT_TIMEOUT = ['instances.create']
+const NO_TIMEOUT_ACTIONS = ['instances.create']
 const ZAF_EVENT = /^zaf\./
 const pendingPromises = {}
 const ids = {}
@@ -26,7 +26,7 @@ function timeoutReject (reject, name, client, paramsArray) {
   switch (name) {
     case 'invoke': {
       const matches = paramsArray.filter((action) => {
-        return PROMISE_DONT_TIMEOUT.indexOf(action) !== -1
+        return NO_TIMEOUT_ACTIONS.indexOf(action) !== -1
       })
       const allWhitelisted = matches.length === paramsArray.length
       if (allWhitelisted) {
@@ -223,7 +223,7 @@ function messageHandler (client, event) {
   }
 }
 
-// When doing singular operations and we retrieve and error this function will throw that error.
+// When doing singular operations and we retrieve an error this function will throw that error
 function processResponse (path, result) {
   const isSingularOperation = typeof path === 'string'
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -5,8 +5,10 @@ import Tracker from './tracker'
 import NativePromise from 'native-promise-only'
 
 const Promise = window.Promise || NativePromise
-const PROMISE_TIMEOUT = 10000
+export const PROMISE_TIMEOUT = 10000
 // 10 seconds, see ZD#4058685
+export const PROMISE_TIMEOUT_LONG = 5 * 60 * 1000
+// Set a high timeout threshold to allow us to better gauge the time taken for successful requests
 const NO_TIMEOUT_ACTIONS = ['instances.create']
 const ZAF_EVENT = /^zaf\./
 const pendingPromises = {}
@@ -21,8 +23,8 @@ window.Promise = Promise
 // Reject a request if required, given the rejection function, name (get, set or invoke)
 // and arguments to that request. Falls back to 10000 second timeout with few exceptions.
 //
-function timeoutReject (reject, name, client, paramsArray) {
-  const actions = paramsArray.map(action => `${name}-${action}`)
+function timeoutReject (client, reject, name, paramsArray) {
+  const actions = collateActions(name, paramsArray)
   switch (name) {
     case 'invoke': {
       const matches = paramsArray.filter((action) => {
@@ -45,6 +47,15 @@ function timeoutReject (reject, name, client, paramsArray) {
   }
 }
 
+function defaultTimer (client, actions, callback) {
+  return setTimeout(() => {
+    trackSDKRequestTimeout(client, actions, PROMISE_TIMEOUT_LONG)
+    callback(new Error('Invocation request timeout'))
+  }, PROMISE_TIMEOUT_LONG)
+}
+
+export function collateActions (name, params) { return params.map(action => `${name}-${action}`) }
+
 export function stripActionArgs (action) {
   // Capture:
   // 1. single or multiple comma-delimited arguments following an initial colon, and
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -7,7 +7,9 @@ import NativePromise from 'native-promise-only'
 const Promise = window.Promise || NativePromise
 const PROMISE_TIMEOUT = 10000
 // 10 seconds, see ZD#4058685
```
Mehul-Porwal: // 5 mins? 🤦‍♂️

fvzzy: This refers to the line above and I've not changed that value. The next line sets the longer timeout actually used in the `setTimeout` function, with an appropriate comment underneath it

lib/client.js
```diff
@@ -131,9 +151,17 @@ function createError (error) {
   return err
 }
 
-function removePromise (id, timeoutId, args) {
+function removePromise (id, timeoutId, client, actions, promiseStartTime, args) {
   clearTimeout(timeoutId)
   delete pendingPromises[id]
+
+  if (promiseStartTime) { // only track time for successfully resolved promises
```
sarmadsangi: ⛏ may be we move this logic to another function so that `removePromise` remains pure?

fvzzy: Not sure I understand. Wouldn't extracting this function (and presumably the additional `client`, `actions`, `promiseStartTime` params) make this _impure_?

lib/client.js
```diff
@@ -119,8 +136,11 @@ function wrappedPostMessage (name, params) {
     rawPostMessage(this, msg)
   })
 
-  // ensure promise is cleaned up when resolved
-  return promise.then(removePromise.bind(null, id, timeoutId), removePromise.bind(null, id, timeoutId))
+  // ensure promise is cleaned up when resolved/rejected, pass the start time for resolved promises
+  return promise.then(
+    removePromise.bind(null, id, timeoutId, this, actions, promiseStartTime),
+    removePromise.bind(null, id, timeoutId)
```
sarmadsangi: Question: not related to this PR, but why do we have to do this?

fvzzy: Which part? It looks as though previously we were passing the same method (with the same params) as both the `resolve` and `reject` callbacks to this promise, though that probably could have been handled via a `finally` (or equivalent) handler instead. Now we're passing additional params when the promise is successfully resolved, and omitting those params in the rejected handler. How might you rewrite this?

sarmadsangi: Ignore this. I totally missed `promise.then` and thought it was `promose.all` and assumed that these are multiple promises.

lib/client.js
sarmadsangi: Yea what I meant is, what if we just move the whole thing out and not even bother passing those arguments.

I might need to spend more time understanding how all the things are hooked up together here but I was thinking something like this would do it;

```
const trackLongRequestsCallback = trackLongRequests.bind(client, actions, promiseStartTime)
promise
  .then(removePromise.bind(null, id, timeoutId), removePromise.bind(null, id, timeoutId))
  .then(trackLongRequestsCallback)
  .catch(trackLongRequestsCallback)
```

`catch` is if we want to catch rejected requests too (which I assume we want to?).

lib/client.js
```diff
@@ -54,18 +65,21 @@ export function stripActionArgs (action) {
...
-    callback(new Error('Invocation request timeout'))
-  }, PROMISE_TIMEOUT)
+function trackSDKRequestTimeout(client, actions, requestResponseTime) {
+  const actionsTags = actions.map(action => `action:${stripActionArgs(action)}`)
+  const responseTimeTag = `request_response_time:${timeToSecondsRange(requestResponseTime)}`
+  client.postMessage('__track__', {
+    event_name: 'sdk_request_timeout',
+    event_type: 'increment',
+    data: 1,
+    tags: [actionsTags, responseTimeTag]
```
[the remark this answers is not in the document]
fvzzy: Yep good catch 👍

lib/client.js
fvzzy: We don't want to catch rejected requests, but otherwise I agree this seems like a better approach 👍

fvzzy: ...actually this won't work since it changes the value returned from `wrappedPostMessage` and  passed to `processResponse` in handlers like `get`, e.g. https://github.com/zendesk/zendesk_app_framework_sdk/pull/88/files#diff-50cfa59973c04321b5da0c6da0fdf4feR596

sarmadsangi: ic, I guess we can just leave this for now then.

lib/client.js
```diff
@@ -131,9 +151,17 @@ function createError (error) {
   return err
 }
 
-function removePromise (id, timeoutId, args) {
+function removePromise (id, timeoutId, client, actions, promiseStartTime, args) {
```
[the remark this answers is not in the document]
fvzzy: For now I'm explicitly passing `null` as the value of these three arguments to get around this in the failed state as I'm not sure of a nicer way to handle it. Does that work?

lib/client.js
fvzzy: @Mehul-Porwal can I resolve this?

lib/client.js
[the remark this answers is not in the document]
hoangtrvu: yup that works. Another way would be to combine these into an object:
````
removePromise.bind(null, { id, timeoutId, client, ...}),
removePromise.bind(null, { id, timeoutId })
// then
function removePromise ({ id, timeoutId, client, ...}, args) { ... }
````

File: spec/client_spec.js [hunk 6]
```diff
@@ -597,10 +637,10 @@ describe('Client', () => {
         }).to.throw(Error)
       })
 
-      it('rejects the promise after 10 seconds', (done) => {
+      it('rejects the promise after the timeout length', (done) => {
```
token-cjg: Should we also include a test to check that it doesn't reject the promise before the timeout length?

lib/client.js
```diff
@@ -102,12 +116,16 @@ function rawPostMessage (client, msg, forceReady) {
...
+
   const promise = new Promise((resolve, reject) => {
-    // Time out the promise to ensure it will be garbage collected if nobody responds
-    timeoutId = timeoutReject(reject, name, this, Array.isArray(params) ? params : Object.keys(params))
+    promiseStartTime = window.performance.now()
 
-    pendingPromises[id] = { resolve: resolve, reject: reject }
+    // Time out the promise to ensure it will be garbage collected if nobody responds
+    timeoutId = timeoutReject(this, reject, name, paramsArray)
+    pendingPromises[id] = { resolve, reject }
```
token-cjg: Forgive me if I'm a bit naff to ask this, but was the alteration of the structure of this object an intentional refactor?  I'm a bit unclear as to how this variable is used.

lib/client.js
```diff
@@ -119,8 +137,11 @@ function wrappedPostMessage (name, params) {
     rawPostMessage(this, msg)
   })
 
-  // ensure promise is cleaned up when resolved
-  return promise.then(removePromise.bind(null, id, timeoutId), removePromise.bind(null, id, timeoutId))
+  // ensure promise is cleaned up when resolved/rejected, pass the start time for resolved promises
+  return promise.then(
+    removePromise.bind(null, id, timeoutId, this, actions, promiseStartTime),
```
token-cjg: Another silly question: how does bind work in this instance?  I see that from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind that `null` would take the roll of `args` in the function itself - is that correct?

lib/client.js
```diff
@@ -131,9 +152,17 @@ function createError (error) {
   return err
 }
 
-function removePromise (id, timeoutId, args) {
+function removePromise (id, timeoutId, client, actions, promiseStartTime, args) {
```
token-cjg: Have unit tests been modified for this altered function?

lib/client.js
sarmadsangi: @token-cjg `pendingPromises[id] = { resolve, reject }` this is es6 object property shorthand.
https://alligator.io/js/object-property-shorthand-es6/

basically, `pendingPromises[id] = { resolve, reject }` is same as `pendingPromises[id] = { resolve: resolve, reject: reject }`

spec/client_spec.js
fvzzy: I think the tests above should cover this as they test the promise is resolved and the appropriate data returned

lib/client.js
fvzzy: It may be easier to explain in person, but in this case we're:
1) keeping the value of `this` when calling this function by setting the first argument to `null`,
2) "pre-setting" the value of the first 5 params in the `removePromise` call with arguments 2-6 in these bound functions,
3) using the returned value from the wrapping `promise.then` call as the remaining `args` when calling `removePromise`.

So when `removePromise` is called it should always have an `args` value depending on whether the wrapping promise was resolved/rejected

token-cjg: Thanks for the detailed and helpful explanation!  I'm still a little bit confused on the first point, maybe we could discuss that later?

lib/client.js
```diff
@@ -102,25 +116,40 @@ function rawPostMessage (client, msg, forceReady) {
...
+  const trackTimeoutWithResolutionTime = () => {
+    const promiseResolutionTime = window.performance.now() - promiseStartTime
+    if (promiseResolutionTime > PROMISE_TIMEOUT) {
+      trackSDKRequestTimeout(client, actions, promiseResolutionTime)
+    }
+  }
+
+  // ensure promise is cleaned up when resolved/rejected, track request performance for resolved promises
+  return promise.then(
+    removePromise.bind(null, { id, timeoutId, trackTimeoutWithResolutionTime }),
```
sarmadsangi: 💯

PR #90, opened 2019-03-26
[AF-1548] [POC] Pause request timeout timers in background tabs

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
1. Adds an event listener to know if a tab changes from foreground <-> background (see: https://developers.google.com/web/updates/2017/03/background_tabs);
2. Pauses/destroys the existing timers on transition to background tab;
3. Resumes/reinitiates the timers if the tab ever resurfaces a foreground tab.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1548

### Risks
* TBD

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1,18 +1,19 @@
 /* global URL */
 import version from 'version'
-import { when, isObject, isString } from './utils'
+import { when, isObject, isString, PausableTimer } from './utils'
 import Tracker from './tracker'
 import NativePromise from 'native-promise-only'
 
 const Promise = window.Promise || NativePromise
 export const PROMISE_TIMEOUT = 10000
 // 10 seconds, see ZD#4058685
-export const PROMISE_TIMEOUT_LONG = 5 * 60 * 1000
+export const PROMISE_TIMEOUT_LONG = 30000
 // Set a high timeout threshold to allow us to better gauge the time taken for successful requests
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -48,10 +49,12 @@ function timeoutReject (client, reject, name, paramsArray) {
 }
 
 function defaultTimer (client, actions, callback) {
-  return setTimeout(() => {
+  const timer = new PausableTimer(() => {
     trackSDKRequestTimeout(client, actions, PROMISE_TIMEOUT_LONG)
     callback(new Error('Invocation request timeout'))
   }, PROMISE_TIMEOUT_LONG)
+  requestTimers.push(timer)
```
fvzzy: Just realised this won't work since the completed timers will need to be removed from this queue or they'll just start all over when we revisit the page a second time - I'll fix this.

lib/utils.js
```diff
@@ -114,3 +114,20 @@ export function when (items) {
 
   return allResolvedPromise
 }
+
+export function PausableTimer (callback, delay) {
```
ocke: I like this function but since it's being used with `new` it could be a class in ES6.

```
export class PausableTimer {
  constructor (callback, delay) {
    ...
  }

  pause () {
    ...
  }

  resume () {
    ...
  }
}
```

PR #95, opened 2019-06-26
[AF-1489] Update Fossa commands

token-cjg:
Let's try to get Fossa builds working properly.

In particular, we'd like to fail the build if there are problems with deep dependencies.


hoangtrvu:
Closing this in favour of https://github.com/zendesk/zendesk_app_framework_sdk/pull/96

PR #96, opened 2019-06-26
[AF-1489] Update Fossa command

hoangtrvu:
:v:

/cc @zendesk/vegemite

### Description
Update Fossa command to run `fossa test` as part of Travis check.

- `fossa` is the shorthand for `fossa init` then `fossa analyze`, which only gather information about dependencies of the project then upload that data to Fossa. This command doesn't run further checks or scans.
- `fossa test` triggers the actual scan based on the data provided by `fossa analyze` in the previous step. We only need to run this if `fossa analyze` completed successfully. This step would take a longer time than `analyze` (default timeout is 10 minutes which is unlikely to be hit based on their documentation) and would report exit status back to Travis once it's done.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1489
* Slack: https://zendesk.slack.com/archives/CHCQYJ468/p1561516891013900

### Risks
* None. This change only affects our internal build pipeline.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,6 @@ matrix:
       before_script:
         - curl -H 'Cache-Control:no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh | BINDIR=$(pwd) bash
       script:
-        - ./fossa # analyze dependencies and upload them (fails when missing api key)
+        - ./fossa && ./fossa test # analyze and scan dependencies (fails when missing api key)
   allow_failures:
     - env: TASK=fossa
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Zendesk App Framework SDK
 =========================
 
 [![Build Status](https://travis-ci.org/zendesk/zendesk_app_framework_sdk.svg?branch=master)](https://travis-ci.org/zendesk/zendesk_app_framework_sdk)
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -22,5 +22,7 @@ matrix:
         - curl -H 'Cache-Control:no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh | BINDIR=$(pwd) bash
       script:
         - ./fossa # analyze dependencies and upload them (fails when missing api key)
+      after_success:
+        - ./fossa test
```
sarmadsangi: Questions:
1. what does this `fossa test` do?
2. why run `after_success` ?

hoangtrvu: I've updated the PR description which would cover these points

sarmadsangi: Don't we want the build to fail if `fossa test` fails? If yes, we might want to move this to run as a 2nd command in `script` instead of `after_success`.

"The exit code of after_success, after_failure, after_script, after_deploy and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as failed."
https://docs.travis-ci.com/user/job-lifecycle/

token-cjg: Yes, I think Sarmad makes a good point.  Note that the overall matrix build will not fail if TASK=fossa fails due to https://github.com/zendesk/zendesk_app_framework_sdk/pull/96/files#diff-354f30a63fb0907d4ad57269548329e3R28 , but we do need to know if the fossa task fails.

hoangtrvu: Hmm strange, Fossa doc recommends exactly this (using `after_success` to fail the build) https://docs.fossa.com/docs/travisci#section-blocking-ci-builds-w-fossa-issue-status. Let me give it a quick test.

hoangtrvu: I verified that exit code is ignored in `after_success`, changed the code to run `fossa test` in `script` now.

PR #97, opened 2019-06-28
Allowlist Zendesk Sell domains

bbrzoska:
✌️

/cc @zendesk/vegemite

## Risks

Low. This affects an informative check that currently displays an error in the console.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -280,8 +280,10 @@ function isOriginValid (origin) {
       /^https:\/\/.+\.zd-dev\.com$/,
       /^https:\/\/.+\.zd-master\.com$/,
       /^https:\/\/.+\.zendesk-staging\.com$/,
+      // Zendesk Chat domains:
       /^https?:\/\/.+\.zopim\.com(:\d+)?$/,
       /^https:\/\/dashboard\.zopim\.org$/,
+      // Zendesk Sell domains:
       /^https:\/\/.+\.futuresimple\.com$/,
       /^https:\/\/.+\.cloudhatchery\.com$/,
       /^https:\/\/.+\.idealwith\.com$/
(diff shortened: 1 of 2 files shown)


Mehul-Porwal:
Do you want to add some unit tests to avoid regressions in future? here: https://github.com/zendesk/zendesk_app_framework_sdk/blob/master/spec/client_spec.js

lib/client.js
```diff
@@ -281,7 +281,10 @@ function isOriginValid (origin) {
       /^https:\/\/.+\.zd-master\.com$/,
       /^https:\/\/.+\.zendesk-staging\.com$/,
       /^https?:\/\/.+\.zopim\.com(:\d+)?$/,
-      /^https:\/\/dashboard\.zopim\.org$/
+      /^https:\/\/dashboard\.zopim\.org$/,
+      /^https:\/\/.+\.futuresimple\.com$/,
+      /^https:\/\/.+\.cloudhatchery\.com$/,
+      /^https:\/\/.+\.idealwith\.com$/
```
sarmadsangi: Just curious why do these domain name sound so unrelated to sell? It would be great to add some details around this in the PR for future references.

sarmadsangi: And also, given that these are not explicit maybe we add a comment block just above these domains stating `Zendesk sell domains`?

bbrzoska: @sarmadsangi it's pretty much the same reason "zopim" is unrelated to Chat (unless I'm missing something 😄) - FutureSimple was the name of the company that built Base, and long time ago the app was setup to run from there (and the cost to switch would be too great, but eventually we'll migrate over to zendesk.com). `cloudhatchery` and `idealwith` are staging and sandbox domains, and I don't think they have any meaning.

We also have a `base.plumbing` domain for some techops stuff - finally that one that means something! :)

I'll add a comment.

Mehul-Porwal:
👍 again!

bbrzoska:
Thank you all for the reviews!

PR #101, opened 2019-08-07
[AF-1697] Only have 1 client

ocke:
Only have 1 client. Return the client if we already had one.

https://zendesk.atlassian.net/browse/AF-1697

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -26,8 +26,11 @@ import { queryParameters } from './utils'
  * ```
  */
 
+let _client
 const ZAFClient = {
   init: function (callback, loc) {
+    if (_client) return _client
+
     loc = loc || window.location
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -26,8 +26,11 @@ import { queryParameters } from './utils'
  * ```
  */
 
+let _client
 const ZAFClient = {
   init: function (callback, loc) {
+    if (_client) return _client
```
hoangtrvu: With this change, we would always return the same client regardless of different `callback` or `loc` being passed, which doesn't look right to me. I've seen apps creating multiple clients in the past during app review process, not entirely sure about their use case but I think this would be a risky change if we suddenly remove this capability.

sarmadsangi:
Closing this, since we don't have immediate plans do make this change.

PR #103, opened 2019-10-14
Update ZAF documentation url

alvesjtiago:
## Description
This is a simple url documentation change since the ZAF link provided was no longer available.


sarmadsangi:
Thanks @alvesjtiago for fixing this.

PR #106, opened 2020-01-29
[AF-1975] Clear iframe when origin isnt whitelisted

ashaban-work:
Stop apps from loading from non-whitelisted domains
Also we unload the app from the DOM if the origin isnt whitelisted

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -299,16 +299,19 @@ function isOriginValid (origin) {
   return false
 }
 
-function clearIframe (origin) {
+function clearDocument () {
   document.body = document.createElement('body')
+
+  const iframeHead = document.head
+  iframeHead && iframeHead.remove()
+}
+
+function appendErrorToBody (origin) {
   var errorElement = document.createElement('h3')
   var errorMsg = document.createTextNode(`Invalid domain ${origin}`)
 
   errorElement.appendChild(errorMsg)
   document.body.appendChild(errorElement)
-
-  const iframeHead = document.head
-  iframeHead && iframeHead.remove()
 }
 
 export default class Client {
@@ -334,7 +337,8 @@ export default class Client {
         tags: ['origin:' + originHostname]
       })
 
-      clearIframe(this._origin)
+      clearDocument()
+      appendErrorToBody(this._origin)
     }
 
     this.on('app.registered', (data) => {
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -299,6 +299,18 @@ function isOriginValid (origin) {
   return false
 }
 
+function clearIframe (origin) {
```
sarmadsangi: ⛏ this could be split into two functions `clearDocument` and `appendTextToBody` (something along this line).

spec/client_spec.js
```diff
@@ -168,8 +168,12 @@ describe('Client', () => {
           appGuid: 'appGuid',
           source: source
         })
+        const ACTUAL_ERROR_MSG = document.querySelector('h3').innerText
+        const EXPECTED_ERROR_MSG = `Invalid domain ${domain}`
+
         expect(client).to.be.an('object')
-        expect(console.error).to.have.been.calledWith('Invalid domain: ' + domain)
+        expect(document.head).to.equal(null)
+        expect(ACTUAL_ERROR_MSG).to.equal(EXPECTED_ERROR_MSG)
```
sarmadsangi: title of this test can be updated now (we don't throw error anymore)

lib/client.js
ashaban-work: Done!

spec/client_spec.js
ashaban-work: yess i forgot about that x_x

lib/client.js
```diff
@@ -322,7 +337,8 @@ export default class Client {
         tags: ['origin:' + originHostname]
       })
 
-      console && console.error('Invalid domain: ' + this._origin)
+      clearDocument()
+      appendErrorToBody(this._origin)
```
sarmadsangi: ⛏ super tiny thought, you can pass full string here instead of origin to make this function generic.

ashaban-work: Done!

spec/client_spec.js
```diff
@@ -148,7 +148,7 @@ describe('Client', () => {
       })
     })
 
-    it('throws when domain is invalid', () => {
+    it('clear iframe and append error msg to body when domain is invalid', () => {
```
token-cjg: nit re proper use of the imperative: `clear` -> `clears`, `append` -> `appends`

ashaban-work: haha yess my grammar is the worst :P

PR #109, opened 2020-02-03
[AF-1975] Clear iframe when origin isnt whitelisted

ashaban-work:
Stop apps from loading from non-whitelisted domains
Also we unload the app from the DOM if the origin isnt whitelisted

### DIFF
--- a/spec/client_spec.js
+++ b/spec/client_spec.js
@@ -168,8 +168,12 @@ describe('Client', () => {
           appGuid: 'appGuid',
           source: source
         })
+        const ACTUAL_ERROR_MSG = document.querySelector('h3').innerText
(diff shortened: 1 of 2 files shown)


spec/client_spec.js
```diff
@@ -148,7 +148,7 @@ describe('Client', () => {
       })
     })
 
-    it('throws when domain is invalid', () => {
+    it('clears iframe and appends error msg to body when domain is invalid', () => {
```
Mehul-Porwal: Can we add one tesco domain to the test case above this?

ashaban-work: Done! should i add one to the invalid domains ?

Mehul-Porwal: nah, that should be alright.

PR #118, opened 2020-07-28
[VEGEMITE-320] Replace Travis with Github actions

anushkafka:
This PR focuses on replacing Travis with Github actions


.github/workflows/actions.yml
```diff
@@ -0,0 +1,34 @@
...
+      matrix:
+        include:
+          - env:
+              TASK: lint
+            allow_failures: false
+            command: npm run lint
+
+          - env:
+              TASK: test
+            allow_failures: false
```
sarmadsangi, line 22: What is the default value of this? sensible default IMO would be false, so if that is the case we can just remove these from here.

cc @shoukoo

.github/workflows/actions.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  main-job:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
```
sarmadsangi, line 12: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

This is set to true by default, we can revisit ZAF and other places where we may have these. Its not a big deal to just leave them there, but for this one we can remove it, since its not merged yet.

.github/workflows/actions.yml
sarmadsangi: Ok had a chat with Andy just now, so `allow_failures` by itself does not do anything at all, its just a variable that we can use it within our workflow like this and it could be named anything since its variable,

```
...
continue-on-error: ${{ matrix.allow_failure }}
strategy:
  fail-fast: false
  matrix:
    include:
      - env:
           TASK: link
        allow_failure: true
....
```

For these tests, we can just remove allow_failure given we always want failed job to exit unsuccessfully. cc @token-cjg

PR #159, opened 2022-02-28
Add a allowOrigin option to bypass invalid origins

luis-almeida:
:v:

/cc @zendesk/vegemite

### Description

In Guide (Help Center) we have successfully built a POC for an enduser app using `zaf`, `zaf_sdk` and `zcli`. We are now doing a final push to release this app in an EAP.

As we talked about in one of our last meetings, Help Center supports host-mapped domains and this is currently leading to `Invalid domain [origin]` error screens. In this PR we're following your recommendation to introduce a way to bypass this check.

We understand this allow list was added in https://github.com/zendesk/zendesk_app_framework_sdk/pull/58 to address a [security vulnerability](https://zendesk.slack.com/archives/C371P4Y3S/p1582088550425200) so I'm thinking that, even thought left undocumented, this solution might reopen that vulnerability.

We'd really appreciate your feedback here 🙏🏼 
Alternatively, help us think of another way to have this available only in Help Center.

### Tasks
- [ ] Add [unit tests](/spec)

### References
* JIRA: https://zendesk.atlassian.net/browse/VEG-XXX

### DevQA Steps

### Risks
* [HIGH | medium | low] Does it work across browsers (including IE!)?
* [HIGH | medium | low] Does it work in the different products (Support, Chat, Connect)?
* [HIGH | medium | low] Are there any performance implications?
* [HIGH | medium | low] Any security risks?
* [HIGH | medium | low] What features does this touch?

### Rollback Plan

1. Notify the author.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -326,6 +326,7 @@ export default class Client {
     this._source = options.source || (this._parent && this._parent._source) || window.parent
     this._appGuid = options.appGuid || (this._parent && this._parent._appGuid)
     this._instanceGuid = options.instanceGuid || this._appGuid
+    this._allowOrigin = options.allowOrigin || false
     this._messageHandlers = {}
     this._repliesPending = {}
     this._instanceClients = {}
@@ -335,7 +336,7 @@ export default class Client {
     this._idleState = null
     this.ready = false
 
-    if (!isOriginValid(this._origin)) {
+    if (!this._allowOrigin && !isOriginValid(this._origin)) {
       const originHostname = new URL(this._origin).hostname
       this.postMessage('__track__', {
         event_name: 'invalid_sdk_origin',
--- a/lib/index.js
+++ b/lib/index.js
@@ -27,15 +27,17 @@ import { queryParameters } from './utils'
  */
 
 const ZAFClient = {
-  init: function (callback, loc) {
+  init: function (callback, loc, options) {
     loc = loc || window.location
+    options = options || {}
+    const allowOrigin = options.allowOrigin
     const queryParams = queryParameters(loc.search)
     const hashParams = queryParameters(loc.hash)
     const origin = queryParams.origin || hashParams.origin
     const appGuid = queryParams.app_guid || hashParams.app_guid
     if (!origin || !appGuid) { return false }
 
-    const client = new Client({ origin, appGuid })
+    const client = new Client({ origin, appGuid, allowOrigin })
 
     if (typeof callback === 'function') {
       client.on('app.registered', callback.bind(client))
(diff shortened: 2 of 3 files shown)


lib/client.js
```diff
@@ -326,6 +326,7 @@ export default class Client {
     this._source = options.source || (this._parent && this._parent._source) || window.parent
     this._appGuid = options.appGuid || (this._parent && this._parent._appGuid)
     this._instanceGuid = options.instanceGuid || this._appGuid
+    this._allowOrigin = options.allowOrigin || false
```
token-cjg, line 329: I suspect that this might not work, because in practice it is the app that initialises the sdk_client [here](https://github.com/zendesk/zendesk_app_framework_sdk/blob/master/lib/index.js#L38).  You would likely need to introduce that as an additional option to be passed either explicitly or implicitly in index.js.

Of course if you did end up implementing this you would be bypassing the whole reason that https://github.com/zendesk/zendesk_app_framework_sdk/pull/58 was done, which was to implement a client-side security feature for the zaf_sdk.  We evidently can't do this, so there would be a need to implement this another way.

I gather that the usecase here is end-user apps, which is moderately interesting.

I gather you're somehow interested in allowing zaf_sdk to pick up a message from our systems and for this you need to work around the allowed origins whitelist.

One potential way to do this might be to write the desired message for zaf_sdk to a pub/sub system eg Kafka or SQS, and then we could have logic in zaf_sdk that polls a whitelisted endpoint in a separate VPC to our other infrastructure (eg like an Untrusted Ingress Zone).  Then somehow we'd like to prove that the request is coming from a legitimate end user app, it seems a bit unclear how to do that ... but assuming that we do have a way to do that, then this nascent ingress service could direct the request to a consumer that listens to the topic, and then the request could be returned to the app via the existing egress lambda service.

Probably for proving that the request is legitimate potentially we could adapt the pattern used in https://github.com/zendesk/zendesk_app_framework_sdk/pull/145 to have a state associated to the app that is a string stored in an object associated to the zaf_sdk, and this is coupled to classic which generates a 1 time password string every `<interval>` eg every 30 seconds, and this could be kept in sync with the consumer in the separate VPC that I've indicated above.

Or we could adopt another approach that is a bit less convoluted but still secure.

But we unfortunately cannot solve your problem via this approach, we'd need to likely adopt a slightly more complex solution.

token-cjg (changes requested):

token-cjg:
I suspect that we'll want to adopt a different approach, as this will create a security issue if we follow it as-is.

GorkaCardona:
Thanks @token-cjg for the fast response! @taoza @dahall-Z3N 

What would be your teams recommendation for a different approach?

PR #175, opened 2022-08-08
[VEG-1404] Bump the version for deployment

etanxing:
:v:

/cc @zendesk/vegemite

### Description
Bump the correct version. https://github.com/zendesk/zendesk_app_framework_sdk/commit/f072478129d5544e8b9c19d9ae31f04158421381 was pushed by accident.

### Tasks
- [ ] Add [unit tests](/spec)

### References
* JIRA: https://zendesk.atlassian.net/browse/VEG-XXX

### DevQA Steps

### Risks
* [HIGH | medium | low] Does it work across browsers (including IE!)?
* [HIGH | medium | low] Does it work in the different products (Support, Chat, Connect)?
* [HIGH | medium | low] Are there any performance implications?
* [HIGH | medium | low] Any security risks?
* [HIGH | medium | low] What features does this touch?

### Rollback Plan

1. Notify the author.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "zendesk_app_framework_sdk",
-  "version": "2.0.37",
+  "version": "2.0.36",
   "main": "build/zaf_sdk.min.js",
   "description": "The Zendesk App Framework (ZAF) SDK is a JavaScript library that simplifies cross-frame communication between iframed apps and ZAF.",
   "homepage": "http://developer.zendesk.com",


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "zendesk_app_framework_sdk",
-  "version": "2.0.37",
+  "version": "2.0.36",
```
zidmus, line 3: It was 37 should be bumped top 38?

etanxing: The latest is v2.0.36 which already has https://github.com/zendesk/zendesk_app_framework_sdk/pull/172 and https://github.com/zendesk/zendesk_app_framework_sdk/pull/174. It was `v2.0.33` and changed to `v2.0.37` by accident.

zidmus (approved):

package.json
token-cjg: We should probably have a process to properly bump the version when we tag zaf_sdk as part of the deployment process.

Maybe we might be able to emulate the way deployments are currently done for gems? This might be worth asking the appropriate internal team about to see if they have plans for a github action along these lines.

token-cjg (approved):

PR #176, opened 2022-08-29
Add package guard GA

zidmus:
:v:

/cc @zendesk/vegemite

### Description
Dependencies which pollute the global in javascript can cause conflicts with versions of libraries used by app developers within their apps. Therefore, to mitigate the risk of conflicts, we would like to identify a way to prevent dependencies from being added to ZAF_SDK unintentionally.

### References
* JIRA: https://zendesk.atlassian.net/browse/VEG-1466

### Risks
* [low] We just changing the GA to prevent adding dependancies

### Rollback Plan

1. Notify the author.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -6,6 +6,21 @@ on:
     - master
 
 jobs:
+  dependency-guard:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Block dependencies from being added to package.json
+        run: |
+          IFS=' ' read -r allowed <<< "native-promise-only zendesk_app_framework_sdk"
+          dependencies=$(npm ls --production --depth=0 --parseable | awk '{gsub(/\/.*\//,"",$1); print}')
+
+          for dep in $dependencies; do
+            if [[ ! ${allowed[*]} =~ (^|[[:space:]])"$dep"($|[[:space:]]) ]]; then
+              echo "::error ZAP SDK does not allow external dependencies, please remove \"${dep}\" dependency from package.json"
+              exit 1
+            fi
+          done
+        shell: bash
   main-job:
     runs-on: ubuntu-latest
     strategy:
--- a/test.sh
+++ b/test.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+set -euo pipefail
+IFS=' ' read -r allowed <<< "native-promise-only zendesk_app_framework_sdk"
+dependencies=$(npm ls --production --depth=0 --parseable | awk '{gsub(/\/.*\//,"",$1); print}')
+
+for dep in $dependencies; do
+  if [[ ! ${allowed[*]} =~ (^|[[:space:]])"$dep"($|[[:space:]]) ]]; then
+    echo "::error ZAP SDK does not allow external dependencies, please remove \"${dep}\" dependency from package.json"
+    exit 1
+  fi
+done
+
+# for dep in $dependencies; do
+#   found="" 
+#   for allow in $allowed[@]; do
+#     if [[ ${allow} == ${dep} ]]; then
+#       found=$allow
+#       continue
+#     fi
+#   done
+#   if [[ -z "$found" ]]; then
+#     echo "::error ZAP SDK does not allow external dependencies, please remove \"${dep}\" dependency from package.json"
+#     exit 1
+#   fi
+# done
+
+
+# for dep in $dependencies; do
+#   found="" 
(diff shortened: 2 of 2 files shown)


.github/workflows/actions.yml
```diff
@@ -6,6 +6,20 @@ on:
...
+  dependency-guard:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Block dependencies from being added to package.json
+        run: |
+          IFS=', ' read -r allowed <<< "native-promise-only, zendesk_app_framework_sdk"
+          dependencies=$(npm ls --production --depth=0 --parseable | awk '{gsub(/\/.*\//,"",$1); print}')
+          
+          for dep in $dependencies; do
+            if [[ ! " ${allowed[*]} " =~ " ${dep} " ]]; then
```
Frederick888, line 18: Do we need to do regexp here?

Frederick888:
Could you run the script through https://www.shellcheck.net/? Ty

.github/workflows/actions.yml
zidmus: I updated the code to do an exact match, otherwise I can replace above with below, but I found above more clear and compact

```
for dep in $dependencies; do
  found="" 
  for allow in $allowed; do if [[ ${allow} == ${dep} ]]; then found=$allow continue fi done
  if [[ -z "$found" ]]; then
    echo "::error ZAP SDK does not allow external dependencies, please remove \"${dep}\" dependency from package.json"
    exit 1
  fi
done
```

zidmus:
> Could you run the script through https://www.shellcheck.net/? Ty

Yep all clean:
image

.github/workflows/actions.yml
Frederick888: How about

```sh
declare -A allowed
allowed=(
  ["native-promise-only"]="1"
  ["zendesk_app_framework_sdk"]="1"
)
declare -a dependencies
IFS=$'\n' read -r -d '' -a dependencies < <(npm ls --production --depth=0 --parseable --silent | awk '{gsub(/\/.*\//,"",$1); print}')

for dep in "${dependencies[@]}"; do
  if [[ -z "${allowed[$dep]}" ]]; then
    echo "::error ZAP SDK does not allow external dependencies, please remove \"${dep}\" dependency from package.json"
    exit 1
  fi
done
```

.github/workflows/actions.yml
```diff
@@ -6,6 +6,20 @@ on:
     - master
 
 jobs:
+  dependency-guard:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Block dependencies from being added to package.json
+        run: |
+          IFS=' ' read -r allowed <<< "native-promise-only zendesk_app_framework_sdk"
```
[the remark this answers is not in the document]
Frederick888: That's largely already done by GHA I believe: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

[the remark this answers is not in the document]
cl8au: In this case, we should define the shell to be bash I believe, since the undefined only uses `bash -e {0}`

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

cl8au:
I guess you need to fulfils the description of the PR :)

.github/workflows/actions.yml
zidmus: To me they are the same, Look I am happy to copy your code in if you insist.

Frederick888: I think they function the same in this particular case however I can spot a few issues in your script that go against general Bash scripting rules. For example `for dep in $dependencies`:

```sh
bash-5.1$ ls
hello.txt       world.txt
bash-5.1$ declare -a foo=('*')
bash-5.1$ for f in $foo; do printf '%s\n' "$f"; done
hello.txt
world.txt
bash-5.1$ for f in "${foo[@]}"; do printf '%s\n' "$f"; done
*
```

I understand this is irrelevant to this particular use case, but I don't see why we shouldn't hold ourselves to the good practice either. But anyway, bottom line is it works, so I'm happy with the script as it currently is.

Frederick888 (approved):

cl8au (approved):

.github/workflows/actions.yml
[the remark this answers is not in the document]
zidmus: 👍

PR #179, opened 2022-09-13
zidan/veg-1466/bump_version_to_2_0_37

zidmus:
:v:

/cc @zendesk/vegemite

### Description
Bump sdk version to 2.0.37

### References
* JIRA: https://zendesk.atlassian.net/browse/VEG-1466

### Risks
* [low] Bump version

### Rollback Plan

1. Notify the author.


txu-gsd (approved):

package.json
```diff
@@ -86,4 +86,4 @@
     ]
   },
   "snyk": true
-}
+}
```
Frederick888, line 89: Any reason to remove this EOL?

zidmus: No, its been removed by the bump scripts/ or the npm lint looks like? 
Lets me test and maybe fix this as well.

Frederick888: If `npm run bump` or `npm run lint` does remove the EOL then let's keep them that way then :)

If they don't then let's bring the EOL back.

Frederick888 (approved):

PR #220, opened 2024-11-04
[RED-2506] Add zaf_sdk user agent to headers for requests

token-cjg:
:v:

/cc @zendesk/vegemite

### Description
- To assist with debugging the source of various requests it would be helpful to indicate whether requests have been touched by the zaf_sdk at any point
- This will help operationally for various things that need to be done to monitor things of an API related nature
- Knowing that a request has been touched by zaf_sdk helps to pinpoint the source + narrow down where things need fixing and/or patching
- It would be good to get the appId too but this is a good first port of call

### Tasks
- [ ] Add [unit tests](/spec)
- [ ] Add label from one of 'patch'/'minor'/'major' if you want to create a release when merged

### References
* JIRA: https://zendesk.atlassian.net/browse/RED-2506

### DevQA Steps

### Risks
* Medium: Touches requests made by the zaf_sdk, but should not incur _too_ much risk as I'm touching the user-agent header which shouldn't be on the critical path.

### Rollback Plan

1. Notify the author.


lib/utils.js
```diff
@@ -1,6 +1,18 @@
 import NativePromise from 'native-promise-only'
+import pkgJson from '../package.json'
+
 const Promise = window.Promise || NativePromise
 
+export function addUserAgent (headers, appId) {
```
caspersg, line 6: You could add a test for this.

caspersg:
FYI: looks like you need to complete the Tasks too.

lib/utils.js
```diff
@@ -1,6 +1,18 @@
...
+  const originalUserAgent = headers && headers['User-Agent'] ? headers['User-Agent'] : ''
```
Shengming-Yan, line 7: nit: Maybe can simplified as `const originalUserAgent = headers?.['User-Agent'] || '';`

lib/utils.js
```diff
@@ -1,6 +1,18 @@
...
+  const zafSdkUserAgentString = `zendesk_app_framework_sdk/sdk_version:${pkgJson.version}/app_id:${appId}`
+  const userAgent = originalUserAgent ? `${originalUserAgent} ${zafSdkUserAgentString}` : zafSdkUserAgentString
```
Shengming-Yan, line 9: nit: consider ```const userAgent = `${originalUserAgent} ${zafSdkUserAgentString}`.trim();```

lib/client.js
```diff
@@ -557,6 +557,9 @@ export default class Client {
         options = { url: options }
       }
 
+      const appId = this._metadata && this._metadata.appId
```
Shengming-Yan, line 560: nit: consider `options.headers = addUserAgent(options.headers, this._metadata?.appId)`

lib/utils.js
Shengming-Yan, line 6: nit: maybe we can call it `updateUserAgentWithAppId` for its purpose? shall we worry the fallback value for appId if it is null

lib/utils.js
```diff
@@ -1,6 +1,18 @@
 import NativePromise from 'native-promise-only'
+import pkgJson from '../package.json'
+
 const Promise = window.Promise || NativePromise
 
+export function addUserAgent (headers, appId) {
+  const originalUserAgent = headers && headers['User-Agent'] ? headers['User-Agent'] : ''
+  const zafSdkUserAgentString = `zendesk_app_framework_sdk/sdk_version:${pkgJson.version}/app_id:${appId}`
```
Shengming-Yan, line 8: I reckon the version in pkgJson doesn't reflect the actual version, this PR didn't touch on the version as well

lib/utils.js
token-cjg: Hi Sheen, thanks for the feedback.  Unfortunately this repo uses node 12, and optional chaining was only released with node 14 ([stack overflow ref](https://stackoverflow.com/a/59574160), [release notes node v14](https://github.com/nodejs/node/blob/fa8f149c0a4f1d7f9b1de64ea2a6e2d2e38143af/doc/changelogs/CHANGELOG_V14.md#update-to-v8-81)).

lib/utils.js
token-cjg: As above, optional chaining is not possible with node 12 =(

lib/client.js
token-cjg: Likewise, node 12 does not support optional chaining 😢

lib/utils.js
token-cjg: Thank you Sheen, should I remove this? Note that I've bumped the version in package.json [in this commit](https://github.com/zendesk/zendesk_app_framework_sdk/pull/220/commits/23e6df883bfab63cfa495c7c8bed55be8b603c91) but I can remove if you prefer.

lib/utils.js
token-cjg: Good idea, done in this commit: https://github.com/zendesk/zendesk_app_framework_sdk/pull/220/commits/23e6df883bfab63cfa495c7c8bed55be8b603c91

token-cjg: Actioned with this commit https://github.com/zendesk/zendesk_app_framework_sdk/pull/220/commits/23e6df883bfab63cfa495c7c8bed55be8b603c91 , thanks for keeping my pull requests honest 😅

lib/utils.js
Shengming-Yan: I reckon when customer refer to zaf_sdk in their app, it is using the [release version](https://github.com/zendesk/zendesk_app_framework_sdk/releases) now. So this version is unrelavant unless we can manually keep align with the release version. 

With regarding to shall we keep this, as reviewing the purpose of this PR, wonder do you need this info for internal debug only? In that case, probably considering adding x-zendesk-app-framework-sdk-version header for this instead of chain the existing user agent? As all `x-zendesk-` headers will be filtered(unless in allowlist) out before zendesk proxy dispatch the request, while in the meantime, internal can still track the zaf-sdk version in the log

